PR #2, opened 2019-01-11
Node.js - use require() instead of fs.read.

dmitrizagidulin:
Current usage of `fs.readFileSync()` means the context cannot be used in the browser (via Webpack), since `fs` is not present.
Using `require()` instead makes it Webpack-compatible.

(This does mean having to rename `security-v1.jsonld` to `security-v1.jsonld.json`, since the Node `require()` engine does not recognize `jsonld` as a valid JSON extension.)


davidlehn:
-1 on `.jsonld.json`.  That just looks silly!  Github pages was serving up .jsonld files with application/ld+json.  I'm assuming that type was due to the extension.  If we must, it could be just `.json`, but please not both together.

On the webpack side, it can be setup to use the json loader for .jsonld extensions.

@dlongley was looking up issues with this approach, I'm not sure what the conclusion was.  There may be some issues with require caching data when we don't really want it to work that way.

Another approach is to use different apis for each environment.  Node can use fs.read, and package.json can override the main file to foo-browser.js that instead uses require.  And if doing that can just use the webpack specific loader syntax so json-loader is used.

davidlehn:
Re not recognizing .jsonld, what use case is that a problem?  For resources served over the web, headers would be used, so not an issue.  If loading locally, you probably know what to do if you are loading such things?  The spec does list .jsonld as an extension, so perhaps any problems will resolve themselves over time?  https://w3c.github.io/json-ld-syntax/#iana-considerations

dmitrizagidulin:
@davidlehn - k, if `.jsonld.json` is too much, then I would advocate for a simple `.json` extension, for these schemas.

mattcollier:
@dmitrizagidulin @davidlehn please continue discussing this until some resolution is reached.

mattcollier:
Tracking: https://github.com/veres-one/did-veres-one/pull/28

dmitrizagidulin:
@davidlehn The benefit to having the contexts use the `.json` extension (instead of `.jsonld`) is that the Node `require()` statement has an automatic JSON loader that comes with it, if the file has that extension. The same is not true for `.jsonld`. So, it's a difference between `require('file.json')` and `JSON.parse(fs.readSync('file.json', 'utf8'))`.

mattcollier:
@dmitrizagidulin @davidlehn @dlongley @msporny @gannan Forward progress is needed here.  Devs have resorted to commiting contexts documents directly into front-end modules because a solution has not been reached here.  This is holding up other work.

It may be the case that our entire practices around developing context related modules need to be reconsidered.  It was acknowledged early on that building this eco-system on `gh-pages` branches might not be a great idea.  We experienced breakage related to this not long ago when a repository was reorganized.  We resorted to that tactic in hopes that we could have a single source or truth related to a given context.

Perhaps it is the case that we must involve some sort of scripting process.  Something along the lines of how the multiformats project updates their tables might work.

https://github.com/multiformats/js-multicodec/blob/master/tools/update-table.py

dlongley:
If there's no argument we can remember against using `require('thejson.json')`, then let's move forward with that approach as it works with node and with webpack. Any objections?

dmitrizagidulin:
No objection from me, I’m a fan of this approach :)

dlongley:
To deal with the `.jsonld` alias problem, we could keep `fs.readFileSync` for node.js and do `require` for webpack with webpack aliases from the `.json` files to their `.jsonld` counterparts. Thoughts?

dlongley:
package.json
```js
{
   /* ... */
   "browser": {
     "index.js": "browser.js",
     "foo-context-v1.json": "foo-context-v1.jsonld",
     /*...*/
   }
}
```

index.js:
```js
  // do require('fs').readFileSync stuff
```

browser.js
```js
  export.foo = require('foo-context-v1.json');
```

dlongley:
I think the above proposal would make it so we can keep `.jsonld` files in the repo so there isn't any MIME type issue -- and we get to load the files properly in node and we can webpack the stuff properly via `.json` aliases. Seems like this solves all the problems. Does it work?

cc: @davidlehn, @gannan08

gannan08:
> I think the above proposal would make it so we can keep `.jsonld` files in the repo so there isn't any MIME type issue -- and we get to load the files properly in node and we can webpack the stuff properly via `.json` aliases. Seems like this solves all the problems. Does it work?
> 
> cc: @davidlehn, @gannan08

This is indeed a viable solution. If we assume that Webpack will always handle the `browser` field properly, then we can expect it to correctly swap out the files needed to run in the browser.

mattcollier:
This was addressed in https://github.com/w3c-dvcg/security-vocab/pull/4

PR #5, opened 2019-04-18
Browser support.

davidlehn:
- Build and distribute static browser version with all contexts.
- Export a `contexts` Map associating ids to contexts.
- Remove exported `v1` and `v2` in favor of new `contexts` Map.

This changes to a standard Map interface exposed as `contexts`.  That would allow get, has, keys, entries, etc.  Using the `contexts` property so the top level could potential include other properties such as the default id.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 3.0.0 - 2019-xx-xx
 
-### Addded
+### Added
 - Build and distribute static browser version with all contexts.
 - Export a `contexts` Map associating ids to contexts.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
 
 ### Added
 - Build and distribute static browser version with all contexts.
-- Export a `contexts` Map associating ids to contexts.
+- Export a `contexts` Map associating context URIs to contexts.
+- Export a `constants` Object associating short ids to contexts URIs.
 
 ### Removed
-- **BREAKING**: Remove exported `v1` and `v2` in favor of new `contexts` Map.
+- **BREAKING**: Remove exported `v1` and `v2` in favor of new `contexts` Map
+  and `constants` Object.
 
 ## 2.0.0 - 2019-03-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # security-context ChangeLog
 
+## 3.0.0 - 2019-xx-xx
+
+### Addded
+- Build and distribute static browser version with all contexts.
+- Export a `contexts` Map associating ids to contexts.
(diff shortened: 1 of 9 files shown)


davidlehn:
This is using a Map so might require a Map polyfill in browser case.  Hopefully top level apps can handle that?

mattcollier:
Very nice.  Looks like even IE 11 has map support, so that doesn't seem like much of a concern: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility

And when we'rd building document loaders we can use this pattern:
```
const documentCache = new Map([...securityContext.contexts, ...foo.contexts, ...bar.contexts]);
```

Unfortunately, this leaves one issue unsolved which is how we implement this here:  https://github.com/digitalbazaar/bedrock-security-context/blob/master/lib/index.js#L10

Can we exports some constants as well?
```
const constants = exports.contants = {};
constants.SECURITY_CONTEXT_V1_URL = 'https://w3id.org/security/v1';
constants.SECURITY_CONTEXT_V2_URL = 'https://w3id.org/security/v2';
```

I think a regular js object would be the right data structure here so that we can use `Object.assign(bedrock.config.constants, securityContext.constants);`  How does that sound?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # security-context ChangeLog
 
+## 3.0.0 - 2019-xx-xx
+
+### Addded
```
dlongley:
```suggestion
### Added
```

davidlehn:
@mattcollier I'm questioning if constants like that really need to be used.  The bedrock based code depends on them at the moment, but is that a generally enough pattern that we should export those symbols from this package?

I do think there is a use case for exporting a default context url.  Along with package semantic versioning that might make code easier to write so you don't have to update "V2" etc.  Though pulling together many of these packages into one would make that versioning a bit tricky.

Maybe it would be good to start simpler and add more exports later as it becomes more clear what the best pattern is?

mattcollier:
If we don't export the default context URL (constants) in some way, we still have to hard code the canonical URL in all the places.  It sounds like we agree that it would help us with Bedrock which I think is exemplary of how anyone would interact with these context modules.  Is there some reason *not* to export the constants as I suggested?

gannan08:
What is the status of this PR?

davidlehn:
Things I'm still unsure about:
- Anyone else have comments on exporting `constants`?  I can't tell if it's a good idea to have them here or not.
- Should a default context be exposed?  Maybe via `constants.SECURITY_CONTEXT_URL`?  Should their be a shortcut for the context data too?
- If constants are exported, should they be explicit with the `CONTEXT` in the name or could that be simplified out?

PR #6, opened 2019-05-01
Export constants.

mattcollier:
I'm uncertain about the change to the webpack config.  I made the change there because without it I was unable to get anything out of the webpack bundle using a simple HTML test I setup, maybe I was just doing it wrong.

https://gist.github.com/mattcollier/a39697a33a2ffe8c3ebed8c0ae59c1cc

Even with the webpack config in this PR, I could not get the `import * as securityvocab from './dis/main.js';` to work as expected.

I was following this guide: https://webpack.js.org/guides/author-libraries/

We *do* want this to be a library right?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
 
 ### Added
 - Build and distribute static browser version with all contexts.
-- Export a `contexts` Map associating ids to contexts.
+- Export a `contexts` Map associating context URIs to contexts.
+- Export a `constants` Object associating short ids to contexts URIs.
 
 ### Removed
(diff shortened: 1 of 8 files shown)


davidlehn:
I tried some more here:
https://github.com/w3c-dvcg/security-vocab/pull/7

I don't think we need an actual library that can be used from a script tag?  If we do, that could be created as an additional build artifact.  I took the approach of supporting CommonJS and ESM.  It started to get messy doing UMD too due to how webpack defaults to the "browser" field before the "module" field.  So it's unclear what you would put in the package file to support all cases.  Seems maybe just moving forward without UMD support would work for this package.

PR #13, opened 2020-01-18
Fix Links

OR13:
Replace dvcg with ccg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,42 +1,41 @@
-Security Vocabulary
-===================
+# Security Vocabulary
 
-This repository contains the [Digital Verification Community Group][DVCG]
+This repository contains the [Credentials Community Group][ccg]
 Security Vocabulary and a [npm package][security-context] that exports related
 contexts and constants.
 
-Security Vocabulary
--------------------
+## Security Vocabulary
 
-Security Vocabulary specification: https://w3c-dvcg.github.io/security-vocab/.
+Security Vocabulary specification: https://w3c-ccg.github.io/security-vocab/.
 
-Context Package
----------------
+## Context Package
 
 The repository contains [JSON-LD][] contexts for the Security Vocabulary.
 These are also packaged in a [npm package][security-context] for CommonJS and
-ES Modules.  To use with NPM and Node.js, use the following:
+ES Modules. To use with NPM and Node.js, use the following:
 
 ```
 npm install security-context
 ```
 
 The package exposes two values:
+
 - `contexts`: A Map from context URI to JSON-LD context.
 - `constants`: An Object of shorthand keys mapped to context URIs.
 
 ```js
-const {contexts, constants} = require('security-context');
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,42 +1,41 @@
...
 ```
 
 The package exposes two values:
+
 - `contexts`: A Map from context URI to JSON-LD context.
 - `constants`: An Object of shorthand keys mapped to context URIs.
 
 ```js
-const {contexts, constants} = require('security-context');
+const { contexts, constants } = require("security-context");
```
msporny: My expectation is that @dlongley @mattcollier @davidlehn and @gannan08 may have issues with the double quotes and the spacing. If you don't touch the code, it'll be easier to merge this.

msporny: My expectation is that @dlongley @mattcollier @davidlehn and @gannan08 may have issues with the double quotes and the spacing. If you don't touch the code, it'll be easier to merge this.

OR13: The only changes are to README.md, package.json, and index.html.

There were no changes to the actual module code itself.

There is no lint config checked in, or test coverage for this module.

We should consider adding those and CI/CD to this repo to ensure uniform contribution to source material.

I have added: 
- https://github.com/w3c-ccg/security-vocab/issues/16
- https://github.com/w3c-ccg/security-vocab/issues/17

msporny: You changed the code in the README.md from what it used to be, granted, white space only, but in a way that is not aligned with all the code samples in all the other repos (at least that we do). Namely, using double quotes instead of single quotes and adding spaces between braces in one line statements. 

This applies to lines 27, 34, 35, and 36. Putting the code sample back would enable me to merge this immediately (which I'd like to do) and then deal w/ the linting stuff at another time.

msporny:
```suggestion
const {contexts, constants} = require('security-context');
```

README.md
```diff
@@ -1,42 +1,41 @@
...
 ```
 
 With ES Modules:
+
 ```js
 // use one of the following forms:
-import * as securityvocab1 from 'security-context';
-import {default as securityvocab2} from 'security-context';
-import {contexts, constants} from 'security-context';
+import * as securityvocab1 from "security-context";
```
msporny:
```suggestion
import * as securityvocab1 from 'security-context';
```

README.md
```diff
@@ -1,42 +1,41 @@
...
+import { default as securityvocab2 } from "security-context";
```
msporny:
```suggestion
import {default as securityvocab2} from 'security-context';
```

README.md
```diff
@@ -1,42 +1,41 @@
...
+import { contexts, constants } from "security-context";
```
msporny:
```suggestion
import {contexts, constants} from 'security-context';
```

msporny:
Fixed code changes, waiting on one more reviewer.

PR #35, opened 2020-05-07
Define publicKeyJwk

OR13:
create `security-v3-unstable.jsonld`

add definition for `publicKeyJwk`

PR #40, opened 2020-06-16
add definition for publicKeyHex based on RFC4648

kdenhartog:
Signed-off-by: Kyle Den Hartog <kyle.denhartog@mattr.global>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1518,7 +1518,7 @@ data that has been abbreviated for the sake of the readability of the example.
       typeof="rdf:Property">
         <h3>publicKeyHex</h3>
         <p>
-A public key Hex property is used to specify the hex-encoded version of the public key, based on <a href="https://tools.ietf.org/html/rfc4648#section-8">section 8 of rfc4648</a>. Hex encoding is also known as Base16 encoding.
+A <code>publicKeyHex</code> property is used to specify the hex-encoded version of the public key, based on <a href="https://tools.ietf.org/html/rfc4648#section-8">section 8 of rfc4648</a>. Hex encoding is also known as Base16 encoding.
         </p>
         <dl>
           <dt>Status</dt>
@@ -1537,7 +1537,7 @@ A public key Hex property is used to specify the hex-encoded version of the publ
     "id": "did:example:123#hexKey",
     "controller": "did:example:123",
     "type": "EcdsaSecp256k1RecoveryMethod2020",
-    "publicKeyHex":"027560af3387d375e3342a6968179ef3c6d04f5d33b2b611cf326d4708badd7770"
+    "publicKeyHex":"027560AF3387D375E3342A6968179EF3C6D04F5D33B2B611CF326D4708BADD7770"
 }
         </pre>
       </section>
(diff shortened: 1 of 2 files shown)


OR13:
We need this to be defined in a JSON-LD Context... since v3-unstable is currently blocked, how can we add things like this... @msporny ?

msporny:
> since v3-unstable is currently blocked, how can we add things like this... @msporny ?

Why is v3-unstable blocked?

kdenhartog:
Yup, that's the part I wasn't sure about if it should be added to v2 or v3. Will add to this PR before merging.

index.html
```diff
@@ -1514,6 +1514,34 @@ <h3>publicKeyPem</h3>
         </pre>
       </section>
 
+      <section id="publicKeyHex" about="https://w3id.org/security#publicKeyHex"
+      typeof="rdf:Property">
+        <h3>publicKeyHex</h3>
+        <p>
+A public key Hex property is used to specify the hex-encoded version of the public key, based on <a href="https://tools.ietf.org/html/rfc4648#section-8">section 8 of rfc4648</a>. Hex encoding is also known as Base16 encoding.
```
msporny, line 1521:
```suggestion
The <code>publicKeyHex</code> property is used to specify the hex-encoded version of the public key, based on <a href="https://tools.ietf.org/html/rfc4648#section-8">section 8 of rfc4648</a>. Hex encoding is also known as Base16 encoding.
```

index.html
```diff
@@ -1514,6 +1514,34 @@ <h3>publicKeyPem</h3>
...
+        <p>
+    The following example demonstrates the expression of a public key
+    in hex format.
+        </p>
+        <pre class="example prettyprint language-jsonld">
+{
+    "id": "did:example:123#hexKey",
+    "controller": "did:example:123",
+    "type": "EcdsaSecp256k1RecoveryMethod2020",
+    "publicKeyHex":"027560af3387d375e3342a6968179ef3c6d04f5d33b2b611cf326d4708badd7770"
```
msporny, line 1540: This example is wrong -- it uses lowercase letters, which the IETF RFC you point to does not. :) ... decisions, decisions. :P 

You should just do what the referred to spec says (use uppercase). There will, inevitably, be people that want to do both, so it'll be up to you to fight that battle and not give in. The easy path will be to support upper and lowercase, and if you do that, you'll destroy the deterministic sorting provided by base32hex.

kdenhartog: Ohhh good catch, let's define it properly here and when I catch people are doing it wrong I'll refer them here to point them in this direction.

OR13: ahhhh! yes this is a great catch... I gotten beaten up by this issue in my NFC implementation.

OR13:
> > since v3-unstable is currently blocked, how can we add things like this... @msporny ?
> 
> Why is v3-unstable blocked?

https://github.com/w3c-ccg/security-vocab/pull/35

I suppose we can define the file again in this PR.

kdenhartog:
> > > since v3-unstable is currently blocked, how can we add things like this... @msporny ?
> > 
> > 
> > Why is v3-unstable blocked?
> 
> #35
> 
> I suppose we can define the file again in this PR.

Yeah I can do it that way and then we'll just need to resolve the merge conflicts on #35 after. At least then we get v3 started as well.

PR #41, opened 2020-06-18
Add JsonWebSignature2020

OR13:
Add defintion for JsonWebSignature2020

we can kick the can of `JsonWebKey2020` vs `JwsVerificationMethod2020` till we have a special topic call.

### DIFF
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -5,9 +5,14 @@
         },
         "https://w3id.org/security/v2",
         {
+            "JsonWebSignature2020": {
+                "@id": "sec:JsonWebSignature2020"
+            },
             "publicKeyHex": {
                 "@id": "sec:publicKeyHex"
             }
         }
     ]
 }
+
+
--- a/index.html
+++ b/index.html
@@ -609,6 +609,77 @@ and verified.
           </pre>
       </section>
 
+      <section id="JsonWebSignature2020">
+
+        <h3>JsonWebSignature2020</h3>
+          <p>
+A Linked Data signature is used for digital signatures on RDF Datasets.
+The default canonicalization mechanism is specified in the RDF Dataset
+Normalization specification, which deterministically names all
+unnamed nodes. The default signature mechanism
+uses a SHA-256 digest and JWS to perform the digital signature.
+          </p>
+          <dl>
+            <dt>Status</dt>
+            <dd property="vs:term_status">stable</dd>
+            <dt>Parent Class</dt>
+            <dd>Signature</dd>
+            <dt>Expected properties</dt>
+            <dd>verificationMethod, jws</dd>
+            <dt>Signature Properties</dt>
+            <dd>
+              <dl>
+                <dt>Default Canonicalization Algorithm</dt>
+                <dd>
+<a rel="canonicalizationAlgorithm"
+  href="https://w3id.org/rdf#URDNA2015">
+https://w3id.org/rdf#URDNA2015</a>
+                </dd>
+                <dt>Default Signature Algorithm</dt>
+                <dd>
+<a rel="signatureAlgorithm"
+  href="https://tools.ietf.org/html/rfc8037#section-3.1">
+  https://tools.ietf.org/html/rfc8037#section-3.1</a>
+                </dd>
+              </dl>
+            </dd>
+          </dl>
+          <p>
+            The JWS used with Linked Data Signatures is the detached and unencoded payload variant. See the links below for details:
+          </p>
+          <ul>
+            <li>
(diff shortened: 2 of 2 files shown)


OR13:
@msporny what can I do to unlock this?

OR13:
@msporny can I merge this?

msporny:
> @msporny can I merge this?

I'm good with it, but you probably also want reviews from at least @dlongley, @davidlehn, and @dmitrizagidulin just to be sure.

contexts/security-v3-unstable.jsonld
```diff
@@ -14,3 +17,5 @@
         }
     ]
 }
+
+
```
dlongley, line 21: Nit: Some unnecessary whitespace added here.

index.html
```diff
@@ -609,6 +609,77 @@ <h3>Ed25519Signature2018</h3>
           </pre>
       </section>
 
+      <section id="JsonWebSignature2020">
```
davidlehn, line 612: Most ids in this doc have microdata.  I assume this was copied from the above which doesn't have that info.

index.html
```diff
@@ -609,6 +609,77 @@ <h3>Ed25519Signature2018</h3>
...
+                <dd>
+<a rel="canonicalizationAlgorithm"
+  href="https://w3id.org/rdf#URDNA2015">
+https://w3id.org/rdf#URDNA2015</a>
+                </dd>
+                <dt>Default Signature Algorithm</dt>
+                <dd>
+<a rel="signatureAlgorithm"
+  href="https://tools.ietf.org/html/rfc8037#section-3.1">
+  https://tools.ietf.org/html/rfc8037#section-3.1</a>
```
davidlehn, line 642: Is the intent to default the same as Ed25519Signature2018 above?  (I'm not entirely clear how this suite works.)

index.html
```diff
@@ -609,6 +609,77 @@ <h3>Ed25519Signature2018</h3>
...
+          </dl>
+          <p>
+            The JWS used with Linked Data Signatures is the detached and unencoded payload variant. See the links below for details:
+          </p>
+          <ul>
+            <li>
+            <a href="https://tools.ietf.org/html/rfc7797">JSON Web Signature (JWS) Unencoded Payload Option</a>
+          </li>
+          <li>
+            <a href="https://medium.com/swlh/json-web-signature-jws-and-jws-detached-for-a-five-year-old-88729b7b1a68">JSON Web Signature (JWS) and JWS Detached for a five-year-old.</a>
```
davidlehn, line 655: I see this is also in Ed25519Signature2018.  Links to actual specs would be better than links to posts for 5 year olds.

index.html
```diff
@@ -609,6 +609,77 @@ <h3>Ed25519Signature2018</h3>
...
+        </ul>
+          <p>
+The example below shows how a basic JSON-LD proof is expressed in a
+JSON-LD snippet. Note that the proof property is directly embedded in the
+object. The signature algorithm specifies how the proof can be generated
+and verified.
+          </p>
+          <pre class="example prettyprint language-jsonld">
+{
+  "@context": ["https://w3id.org/security/v1", "http://json-ld.org/contexts/person.jsonld"]
```
dlongley, line 666: This example is using an `@context` (security/v1) that doesn't define `JsonWebSignature2020`. I'm not sure if any additional prose should highlight that fact or if a better example could be used. Just making a note of this, I won't hold up merging this PR on this basis, I wouldn't be surprised if other examples also need revamping.

OR13: I opened this issue, we should update them all together: https://github.com/w3c-ccg/security-vocab/issues/48

can you leave a comment there on how you want the updates to look?

I can take a stab at it with more info.

index.html
OR13: The link to the spec is above.... I have yet to encounter a single developer who was not confused by rfc7797... its obscure in the JOSE ecosystem, if you have a better friendly explainer, I'd be in favor of adding another link / replacing this one...

index.html
OR13: This suite is functionally identical to both `Ed25519Signature2018` and `EcdsaSecp256k1Signature2019`... both generate create Verity Data and then use a detached jws with special header values to sign it, and both use `alg` `EdDSA` and `ES256K` respectively... the difference is that this suite also supports all the NIST curves (which some of use are forced to use).... and it relies on this table here for providing guidance on other JOSE suites: https://github.com/w3c-ccg/lds-jws2020#supported-jose-algorithms

index.html
OR13: would you mind commenting on this issue so I can address all these cases at once: https://github.com/w3c-ccg/security-vocab/issues/49

davidlehn: Not sure what to comment on.  The title seems good enough.

OR13: leave advice on the issue so that someone less familar with microdata can take a stab a resolving it. I don't know how to address your concern based on your comments here.

Issue #44, opened 2020-07-02
Define sameAs


OR13:
Useful for describing that 2 subjects are the same entity (id field associated with a verification method)... as discussed at the F2F yesterday.

We should define it here, in the did spec registries and in core with normative text.

dlongley:
To be clear, this would be for describing that 2 *subjects* are the same.

OR13:
Yes, I was trying to make it clear that his would also apply to linked data signatures that do not use DIDs.... one of the most valuable use cases for this is https://auditor.example.com/inspector/123 === did:example:456.

dlongley:
+1 ... I don't think "subjects" or "controllers" are necessarily DID-specific (they are in DID land of course), but I think "sameAs" is a property that would appear on a subject (whether it is identified by a DID or non-DID URI) -- and link it to another subject (again, identified by a DID or non-DID URI). And, I believe, it's still an open question as to whether or not we need a property and definition separate from "owl:sameAs" or if we can just use that vocabulary.

PR #45, opened 2020-07-02
Add sameAs

OR13:
Add sameAs to security vocab.

Addresses: https://github.com/w3c-ccg/security-vocab/issues/44

Cross linking: 

https://github.com/w3c/did-core/issues/33

https://github.com/w3c/did-spec-registries/issues/80

### DIFF
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -13,7 +13,7 @@
                 "@id": "sec:publicKeyHex"
             },
             "sameAs": {
-                "@id": "owl:sameAs-def"
+                "@id": "owl:sameAs"
             }
         }
     ]
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -5,11 +5,15 @@
         },
         "https://w3id.org/security/v2",
         {
-            "ethereumAddress" {
+            "owl": "https://www.w3.org/TR/owl-ref/#",
+            "ethereumAddress": {
                 "@id": "sec:ethereumAddress"
             },
             "publicKeyHex": {
                 "@id": "sec:publicKeyHex"
+            },
+            "sameAs": {
+                "@id": "owl:sameAs-def"
             }
         }
     ]
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -5,7 +5,7 @@
         },
         "https://w3id.org/security/v2",
         {
-            "owl": "https://www.w3.org/TR/owl-ref/#",
+            "owl": "http://www.w3.org/2002/07/owl#",
             "ethereumAddress": {
                 "@id": "sec:ethereumAddress"
             },
--- a/index.html
+++ b/index.html
@@ -1099,6 +1099,31 @@ algorithm:
         </pre>
       </section>
 
+      <section id="sameAs" about="https://w3id.org/security#sameAs"
+        typeof="rdf:Property">
+        <h3>sameAs</h3>
+        <p>
+          owl:sameAs is used to state that two URI references refer to the same individual. 
+          See also <a href="https://www.w3.org/TR/owl-ref">https://www.w3.org/TR/owl-ref</a>
+        </p>
+        <dl>
+          <dt>Status</dt>
+          <dd property="vs:term_status">stable</dd>
+          <dt>Range</dt>
+          <dd>IRI</dd>
+        </dl>
+        <p>
+The example below shows how to express that two subjects are the same, using the OWL sameAs relation.
+        </p>
+        <pre class="example prettyprint language-jsonld">
+{
+  "@context": ["https://www.w3.org/ns/did/v1"],
+  "id": "did:example:123",
+  "sameAs": "https://auditor.example.com/inspector/456",
+}
+        </pre>
+      </section>
+
       <section id="owner" about="https://w3id.org/security#owner"
         typeof="rdf:Property">
         <h3>owner</h3>


contexts/security-v3-unstable.jsonld
```diff
@@ -5,11 +5,15 @@
...
         {
-            "ethereumAddress" {
+            "owl": "https://www.w3.org/TR/owl-ref/#",
+            "ethereumAddress": {
                 "@id": "sec:ethereumAddress"
             },
             "publicKeyHex": {
                 "@id": "sec:publicKeyHex"
+            },
+            "sameAs": {
```
msporny, line 15: Note that @talltree is currently adding this same property to the DID Core specification. Having it in the security vocabulary is dangerous because some developers might think it could be used to assert that a publicKey is the sameAs another public key. sameAs is an incredibly dangerous property in OWL land... it means that two things are semantically equivalent. So, if I were to say #orie sameAs #manu -- I would say that Orie is *exactly the same thing* as Manu, which is problematic, from a DID perspective. We are probably going to want to dial back the OWL semantics, possibly using something else to avoid the sledgehammer that is sameAs. We probably want something closer to "supersededBy" or "replacedBy" to match the use case that I think the community wants.

-0.8 to this PR until we can ferret out the use case.

/cc @iherman @dlongley

OR13: well, its not assuming they are equivalent... the controller is saying the subjects are equivalent... if we don't trust the controller to make assertions about the subject... I fear we have made a mistake :)

Another thing to consider is that the strength of all RDF statements related to a subject is a function of the strength of the DID Method... so for example anyone with push access can add sameAs here: https://github.com/transmute-industries/did.actor/blob/master/alice/did.json#L3 ... whereas in did:v1, you would need private keys... however, assuming private key compromise is in some ways similar to assuming XSS... once its assumed, there are a whole bunch of other security issues... but in my opinion all of them flow from the XSS... similarly for private key compromise... the only reason you trust ANY assertion about the subject is that you believe that private key compromise HAS NOT occurred... once thats inverted, all the assertions (including sameAs) are worthless.

There is advantage, and danger is using OWL directly... but one thing for sure is that using OWL allows for OWL, whereas not using OWL is making something up that is new / not in use / not standardized...

iherman: > Note that @talltree is currently adding this same property to the DID Core specification. Having it in the security vocabulary is dangerous because some developers might think it could be used to assert that a publicKey is the sameAs another public key (suggesting that all associated attributes can be merged together). [...]
> 
> To put it more concretely, one attack that sameAs enables is allowing the merging of compromised keys from an old DID to a new DID... because they're the same thing, right? (except they're not, and assuming they are opens us up to attack vectors).
> 
> -0.8 to this PR until we can ferret out the use case.
> 
> /cc @iherman @dlongley

I am not a security expert but I share the concern of @msporny. For example:

> well, its not assuming they are equivalent... the controller is saying the subjects are equivalent...  (@OR13)

Yes, we can/should trust the controller. The issue is that RDF/OWL is based on an open world assumption; this means that a `owl:sameAs` statement can be made by _anybody in the world_, it does not have to be the controller. It can be you, Manu, or me publishing this statement on the open Web. A bona fide OWL reasoning process that gets hold of this extra triple can make deductions that we may not want. 

As long as we remain within the DID world we are shielded from this because a DID processor is not expected to do any kind of OWL processing. I.e., this does not affect us. Introducing `owl:sameAs` _into_ the DID world gets us closer to OWL, even if minimally, and this may have unexpected consequences that we may not fully understand right now. I would just prefer to be cautious.

As far as I can see the goal is to use `owl:sameAs` in a very very constrained environment, and I personally would be o.k. to introduce our very own `identicalKey` term (or whatever purposes we want to use `sameAs`). Our vocabulary is small, and the places where an identification may be necessary is even smaller, I would not think this would be very onerous.

contexts/security-v3-unstable.jsonld
```diff
@@ -5,11 +5,15 @@
         },
         "https://w3id.org/security/v2",
         {
-            "ethereumAddress" {
+            "owl": "https://www.w3.org/TR/owl-ref/#",
```
peacekeeper, line 8:
```suggestion
            "owl": "http://www.w3.org/2002/07/owl#",
```

contexts/security-v3-unstable.jsonld
```diff
@@ -5,11 +5,15 @@
...
+            "ethereumAddress": {
                 "@id": "sec:ethereumAddress"
             },
             "publicKeyHex": {
                 "@id": "sec:publicKeyHex"
+            },
+            "sameAs": {
+                "@id": "owl:sameAs-def"
```
peacekeeper, line 16:
```suggestion
                "@id": "owl:sameAs"
```

msporny:
@OR13 wrote:
> If we don't trust the controller to make assertions about the subject... I fear we have made a mistake

That's not the point of contention... I don't think anyone is asserting that. The issue is as Ivan says:

@iherman wrote:
> A bona fide OWL reasoning process that gets hold of this extra triple can make deductions that we may not want.

This is the attack. Using owl:sameAs will cause reasoners to treat one DID as *exactly the same thing as* some other DID. That's not the use case and that's not what we want, not only because it'll lead to strange graph merging, but because those graph merges could result in terrible security vulnerabilities that we have no control over.

The solution is as Ivan states... be surgical about the property we're picking. I fear that we're defining a new thing, which is fine, but this property has some pretty dire security concerns and so we need to be very precise with the term we use. It /may/ be a security vocab thing, in fact, it probably is... but we need to constrain it so that OWL reasoners don't go haywire and unwittingly open up all sorts of key aliasing attacks.

contexts/security-v3-unstable.jsonld
iherman: Yes, `http://www.w3.org/2002/07/owl#` is indeed the right OWL URL.

OR13:
> @OR13 wrote:
> 
> > If we don't trust the controller to make assertions about the subject... I fear we have made a mistake
> 
> That's not the point of contention... I don't think anyone is asserting that. The issue is as Ivan says:
> 
> @iherman wrote:
> 
> > A bona fide OWL reasoning process that gets hold of this extra triple can make deductions that we may not want.
> 
> This is the attack. Using owl:sameAs will cause reasoners to treat one DID as _exactly the same thing as_ some other DID. That's not the use case and that's not what we want, not only because it'll lead to strange graph merging, but because those graph merges could result in terrible security vulnerabilities that we have no control over.
> 
> The solution is as Ivan states... be surgical about the property we're picking. I fear that we're defining a new thing, which is fine, but this property has some pretty dire security concerns and so we need to be very precise with the term we use. It /may/ be a security vocab thing, in fact, it probably is... but we need to constrain it so that OWL reasoners don't go haywire and unwittingly open up all sorts of key aliasing attacks.

So in the case that I am already using OWL-RL and doing forward chain reasoning... you are suggesting that we prevent the reasoner form using owl:sameAs... because it might interpret it... correctly?

I'm confused here...  we're starting to have security conversations about semantic reasoners... which like any reasoning system are function of their input.

If you feed a prolog program bad facts and rules... you get bad reasoning... if you feed an OWL-RL system triples containing sameAs (which anyone can construct), and they are not correct... you get bad reasoning...

What we need is layers here:

Layer 1 - Vocabulary and Definitions
Layer 2 - Semantic Assertions using Cryptography
Layer 3 - Semantic Reasoning

Imagine we remove layer 2.... the problem persist... You can still feed `cat1 sameAs dog1` to any reasoner... and it will still lead to problems... its actually worse when there is no ability to authenticate the triple.

creating a new term does not solve the problem... because you can craft `cat1 identicalTo dog1` just as easily....

We should be very careful creating duplicate terms to avoid conflicts with semantic reasoning systems... some people want to use those systems directly.... A system managing aliases for spies, or protecting the privacy of individuals who use multiple dids to defend against correlation might want to reason about the fact that `did1 sameAs did2` is actually a true statement.

dlongley:
I'm so far in agreement with @OR13, particularly with his last comment. Whether or not particular assertions are to be trusted is a separate layer from reasoning about them. I don't understand why the concern is that someone may make a statement that is untrusted -- that has always been possible. I also realize that we'd want bidirectional assertions to trust `sameAs` assertions about DIDs ... but, if we had that, why wouldn't we want to enable reuse of existing reasoners? If we invent a new "sameAs2" -- won't reasoners just have to be updated to support that?

I feel like there's a layering violation in the arguments I've seen so far against reusing "owl:sameAs". If reusing "sameAs" is not semantically correct for some reason, that's a good argument for a new term, but I can't see how whether or not we trust such an assertion is.

David-Chadwick:
+1

wyc:
> ALTHOUGH. I'd much prefer it if we _required_ it to be bi-directional (or at least, made that a part of the definition).

I can imagine instances where it must be unidirectional, especially when the entity in reference is not the same as the DID controllers. For example, two investigators might have their own private internal representations of entity `E` via its public URI, and they may wish to trade their understanding with each other without the involvement of entity `E`.

msporny:
> If reusing "sameAs" is not semantically correct for some reason, that's a good argument for a new term

Reusing sameAs is not semantically correct -- we are not asserting that one DID is exactly the same as some other DID and you should merge their graphs together. We are stating that one DID is asserting that another DID supersedes it. At least, that's the use case I'm talking about. As I said above:

> -0.8 to this PR until we can ferret out the use case.

So, what's the use case? I thought it was: "I was using DID X, and now I'm using DID Y and I want to note in DID X that people should use DID Y from now on." -- also, this creates havoc wrt. reasoning about VCs. Can DID Y provide VCs that were issued to DID X but sign them as DID Y? Doesn't this make the whole ecosystem more complex? Yes, we could use ZKPs for ABCs here, but remember that there are deep concerns about that approach.

We need a list of use cases that "sameAs" and "supersededBy" are supposed to solve. Hard to design a feature without those laid out. We're bound to miscommunicate until they are.

dmitrizagidulin:
@msporny 
> We need a list of use cases that "sameAs" and "supersededBy" are supposed to solve. 

The use cases are as follows.

`sameAs`: One DID is asserted to be equivalent as another by the DID controller. Importantly, both DIDs are available / resolvable, and either can be used interchangeably (so, for example, if they're being used for an identity-based access control system, authenticating as did A is equivalent to authenticating to the sameAs DID B). Bi-directional references are required for validation.

`supersededBy`: The controller is migrating (or has migrated) from DID A to DID B. (DID A is being superseded by DID B). In this case, only DID B needs to be resolvable (in case DID A went away). Although bi-directional reference is preferred, if the "source" DID A is missing, this needs to be paired with a Verifiable Credential (signed by DID A), claiming that DID B is indeed superseded.

Typing this out, this very much seems that this PR should be to the DID Context, instead of the Security context, actually.

iherman:
> `sameAs`: One DID is asserted to be equivalent as another by the DID controller. Importantly, both DIDs are available / resolvable, and either can be used interchangeably (so, for example, if they're being used for an identity-based access control system, authenticating as did A is equivalent to authenticating to the sameAs DID B). Bi-directional references are required for validation.

Thanks for the example. A question: does it mean that the statements of the DID documents become all interchangeable? I.e., let us say the two DID-s are `D_1` and `D_2`, with `Doc_1` and `Doc_2` as DID documents, respectively. `Doc_1` defines, say, a service via the `service` predicate; are we sure that this service is valid for `D_2`, too (because that is what `owl:sameAs` would mean).

Another question, but from another point of view. `owl:sameAs` is, strictly on a fix URI. So if we say `D1 owl:sameAs D2`, that statement does not have _any_ bearing on the respective DID URLs. I.e., if `Doc_1` has a statement on, say, `D_1#123`, that does _**not**_ mean that `D_1#123 owl:sameAs D_2#123`. Is that o.k., or may that be a problem?

OR13:
@dmitrizagidulin we need to keep security vocab in sync... and its not going to get accepted into did spec registries if its not defined somewhere in sufficient detail... 

we need sameAs defined in 3 places... here, did spec registries, and did core.

... in that order.

msporny:
@dmitrizagidulin -- thanks for the use cases, now we just need all the ones from @OR13, @dlongley, @David-Chadwick, and at least everyone else in this thread.

> Importantly: it DOES NOT imply that the graphs for the two DIDs can be merged.

Well, that's one of the points of contention, because that's exactly what owl:sameAs implies. So, owl:sameAs specifically does not solve the use case you stated because of the restriction made in your last sentence, right?

David-Chadwick:
thinking more deeply about it, one DID is not the same as another DID 
even if both belong to the same controller. The controller is the same 
but the way of authenticating him/her/it is different. So maybe a 
different verb is needed such as belongsToSameAs

kind regards

David

dmitrizagidulin:
@msporny 
> So, owl:sameAs specifically does not solve the use case you stated because of the restriction made in your last sentence, right?

Hmm, does seem that way. Maybe need another term. `did:sameAs`.

dlongley:
> Hmm, does seem that way. Maybe need another term. did:sameAs.

If we find that the semantics *are* different (between what we want and `owl:sameAs`), let's not use the term `sameAs` at all, but something that would help make the semantics more clear. I'm not quite sure what we want -- and it does seem like we have more than one use case, or at least, @dmitrizagidulin has highlighted two different ones.

In the first use case, it seems like we want to be able to say there's an alternate DID that could be used to refer to the same subject. Sort of like "Subject Alternative Name" from X.509. It seems to just be informational and doesn't imply that one DID should be preferred over the other for any reason. Also, the "use case" is described more like a feature than like a use case ... so what's this feature for? Do we need to use the DID Document for it -- or could this be done another way ... and is that other way more appropriate?

In the second case, it seems there's a desire to sort of "decommission" an old DID and move to a new one. Again, the "use case" is described as a feature, but it's a bit more clear to me what a possible use case would be and why this could appear in a DID Document. Let's say that Pat has been using the DID `did:example1:1234` to represent themselves on social media. They have built up some relationships using the old DID but now they want to switch to a new DID method "example2". They create a new DID `did:example2:4567` but they don't want to have to notify everyone individually about the fact that they are decommissioning their old DID and adopting this new one as a replacement.

So, Pat uses the `replacedBy` (bikeshed) feature by adding `"replacedBy": "did:example2:4567` to the DID Document for `did:example1:1234`. Now anyone resolving `did:example1:1234` can see this information and use it to discover Pat's new DID.

tplooker:
I would like to see a term that is not constrained to only equating did's, instead one being able to link say a did to a domain name is also very useful, such as described here https://identity.foundation/specs/did-configuration/

dmitrizagidulin:
What if we used ActivityPub's `alsoKnownAs` term? That spec specifically uses it for moving profiles (or linking equivalent ones).

OR13:
Lets separate the desire to use owl:sameAs and the desire to use other identifiers... this PR is for those who wish to use owl:sameAs with existing semantic reasoners... so we don't need to modify them, here is a list of such reasoners:

https://github.com/semantalytics/awesome-semantic-web#reasoners

- [Pellet](https://github.com/stardog-union/pellet)
- [openllet](https://github.com/Galigator/openllet)
- [FaCT++](https://github.com/ethz-asl/libfactplusplus)
- [HermiT](http://www.hermit-reasoner.com/)
- [ELK](https://github.com/liveontologies/elk-reasoner)
- [Whelk](https://github.com/balhoff/whelk)
- [OWL-RL](https://github.com/RDFLib/OWL-RL)
- [RacerPro](https://franz.com/agraph/racer/)
- [Manchester List of Reasoners](http://owl.cs.manchester.ac.uk/tools/list-of-reasoners/)
- [elephant-reasoner](https://github.com/sertkaya/elephant-reasoner)
- [HyLAR](https://github.com/ucbl/HyLAR-Reasoner)
- [ruby-rdf/rdf-reasoner](https://github.com/ruby-rdf/rdf-reasoner)
- [cel](https://github.com/julianmendez/cel) -A lightweight Description Logic reasoner for large-scale biomedical ontologies.
- [reasonable](https://github.com/gtfierro/reasonable) - OWL 2 Reasoner built on DataFrog

If your use case is "I want to not use semantic reasoners but also indicate some kind of equivalence"... you should open another pull request, and explain your use case, and provide vocabulary defintion for it....

msporny:
> This PR is for those who wish to use owl:sameAs with existing semantic reasoners.

Oh, then that makes this PR easy :) -- I would object to any use of owl:sameAs for at least the following two reasons:

* owl:sameAs is a foot gun when combined with cryptography/security... asserting equivalence between any two direct or linked cryptographic identifiers is just asking for a world of pain security vulnerabilities. Introducing this would violate a key lemma in cryptography -- cryptographic identifiers are unique.
* It takes us squarely into non-pragmatic, theoretical, academic ivory tower discussions about semantic reasoning... and we shouldn't waste the community's time with those discussions. Those discussions harmed the adoption of Linked Data for over a decade and many of us don't want to go back into the RDF dark ages.

I find it highly suspect that people are using semantic reasoners in anything other than an academic or experimental system. If people *are* doing that, I'd strongly advise them to stop playing with fire (owl:sameAs when applied to cryptographic primitives). If they continue to play with that fire, let's let them burn their lab down to the grown and not take the rest of us with it.

index.html
```diff
@@ -1148,6 +1148,31 @@ <h3>canonicalizationAlgorithm</h3>
         </pre>
       </section>
 
+      <section id="sameAs" about="https://w3id.org/security#sameAs"
```
davidlehn, line 1151: Not sure what's going on here.  The property in the context is `owl:sameAs`, not `sec:sameAs`.

index.html
```diff
@@ -1148,6 +1148,31 @@ <h3>canonicalizationAlgorithm</h3>
...
+        typeof="rdf:Property">
+        <h3>sameAs</h3>
+        <p>
+          owl:sameAs is used to state that two URI references refer to the same individual.
```
davidlehn, line 1155: Should "individual" be "subject"?

talltree:
I was not aware of this thread when I posted [this PR for a sameAs property in DID Core](https://github.com/w3c/did-core/pull/348). Thanks to @msporny for pointing it out.

I believe I understand @msporny 's concerns. While I also understand @OR13 's desires to pursue full owl:sameAs semantics, the purpose of my [DID Core PR](https://github.com/w3c/did-core/pull/348) (despite the proposed `sameAs` name) was much simpler: just to establish a standard way for two or more DID controllers to assert that their DIDs _identify the same DID subject_. Full stop. No additional semantics or DID document equivalence or RDF graph merging implied.

I'll make that same comment on my DID Core PR. (It would be nice if we could have this conversation in just one place or the other, but not sure what to do about that.)

ChristopherA:
One of my architectural problems with sameAs in DIDs is that in order for it to be provably correct, it needs to be at least an authenticated two-way pointer with another sameAs pointing back to the original DID, and ideally with key material in common. Especially as many DID methods do not individually sign DID Documents, or use different key material, this become problematic.

This is not unrelated to a problem that I also have with VCs where sometimes claim data needs to refer to information that is in the proof part of the claim but is not part of the claim itself, in particular data that is in signatures and/or timestamped proof of knowledge and proof of existence.

For instance, one VC I want soon so that BTCR can replace PGP, is a you "know" the other party because you've done an interactive proof of control of the private key. @dlongley has said this is possible with VCs in issue https://github.com/WebOfTrustInfo/btcr-hackathon-2019/issues/12#issuecomment-520047418 as an "evidence" field in the claim data portion, but I've never seen a working example of what this actually would look like in a VC.

This also brings other architectural questions.

Shouldn't sameAs really be done as a VC instead? Since it can't truly be trusted in a DID without additional trust building/evaluation as per @talltree PR https://github.com/w3c/did-core/pull/348 then I suspect it requires the same trust building/evalutation process that VCs must use. So why have it in the DID document itself?

Which then makes we wonder what other things in DID Documents should be some kind of VC that is attached to a "DID Presentation" along with the base DID Document. For instance the always controversial "service endpoint". If it was a a VC that accompanies a DID Document in a specific DID Presentation between two parties, many of my correlation and privacy concerns about service endpoints would be addressed as DID Presentations can be iterative (aka the holder present a raw DID, the receiving party says "not enough, now send me the DID Document", it is sent and now the receiving party said "I need more" and the holder then choose not to send it for privacy reasons.)

-- Christopher Allen

OR13:
Sadly we are going to have versions of this conversation everywhere, since DID Core JSON-LD context relies on the security vocab, which is hosted here.

As I noted on the PR... saying something and proving something and something being true are all different.

did:example:123 is a dog
did:example:123 self certifies to dogness
did:example:456 certifies did:example:123 is a dog
(did:example:123 is in fact a cat)...

I'm glad we are having this debate, because its fundamental to the work we are all doing.

The point of languages is to express information, and semantic technologies like OWL / RDF allow for unambiguous communication of concepts that can be interpreted by machines.... but critically, they don't tell you anything about the truth... 

they tell you about how a vocabulary term is defined and used consistently... but even then, there is opportunity for subversion... `dog` is `spy`, `walking the dog` is `dead drop`... `feeding the dog` is `confirmation`, etc... anyone can remap a vocabulary and create a parallel narrative, based on an existing set of vocabulary terms... 

adding digital signatures and self certifying identifiers to this doesn't change a thing... except that now, if you trust that some identifier is not compromised, you can trust that statement that are signed by it, came from it.

what the statement actually means is still defined by the vocabulary... or the shadow vocabulary....

and reasons you trust statements from that identifier are still not anyone's business but yours :)

OR13:
Related comment: https://github.com/WebOfTrustInfo/btcr-hackathon-2019/issues/12#issuecomment-659014596

dmitrizagidulin:
@OR13 
> saying something and proving something and something being true are all different.

All very good points. However, the issue that @msporny brought up still stands - `owl:sameAs` and cryptography just do not mix.

OR13:
> @OR13
> 
> > saying something and proving something and something being true are all different.
> 
> All very good points. However, the issue that @msporny brought up still stands - `owl:sameAs` and cryptography just do not mix.

@dmitrizagidulin saying things do not mix is not constructive, or collaborative. 

here is some evidence that they do mix:

- https://ieeexplore.ieee.org/document/5375927
- http://ontology.universaal.org/Cryptographic.owl
- http://www.ai.sri.com/daml/services/owl-s/security.html
- http://www.ijesird.com/march_2.PDF

Share with us some evidence to support your position please.

index.html
OR13: https://www.w3.org/TR/owl-ref/#sameAs-def ... 

> Such an owl:sameAs statement indicates that two URI references actually refer to the same thing: the individuals have the same "identity".

index.html
OR13: I suppose this is an artifact of how https://w3id.org/security handles vocabulary definitions.

Because of the way that the VC Data Model, DID Core Spec and DID Spec Registries rely on this CCG repo for terminology definitions.... I am attempting to define the way sameAs is used in Linked Data Security vocabulary, so that everything is consistent.

If I don't than the behavior of JSON-LD processing software, such as vc.js and jsonld-signatures will be to ignore these properties... 

Can you make a proposed change to ensure that such a thing would not happen?

iherman:
>> All very good points. However, the issue that @msporny brought up still stands - owl:sameAs and cryptography just do not mix.
>
> @dmitrizagidulin saying things do not mix is not constructive, or collaborative.
>
> here is some evidence that they do mix:
> 
> * https://ieeexplore.ieee.org/document/5375927
> * http://ontology.universaal.org/Cryptographic.owl
> * http://www.ai.sri.com/daml/services/owl-s/security.html
> * http://www.ijesird.com/march_2.PDF
> 
> Share with us some evidence to support your position please.

I had a peek into these because this triggered my curiosity, although (not having an IEEE account) I could not read the first one. But these references do not seem to hold as arguments pro on con I am afraid. 

The issue is not whether _OWL_ and crypto (or, to be more specific, DID) mix or not. They can mix as those examples show. But if one looks at the details, the second entry is very instructive. It is an OWL ontology that _describes a vocabulary_ in OWL. It contains a series of definitions for functional properties whose domain and range are also specified by the ontology. This means that OWL reasoners can look at a specific data, use the OWL ontology, and can detect, for example, inconsistencies in the data or deduce the type of a resource. They can also _deduce_  types of resources (based on, say, the domain for a particular property. Mind you, those OWL vocabularies _cannot_ be used to check the _validity_ of the data; OWL is not equipped to do that.

We could, for example, decide to create such an ontology for the DID vocabulary, although I am not sure it would be really useful and/or widely used (OWL is really useful when one has very large graphs, and a DID document is really tiny).

I think the statement of @msporny is not whether OWL and crypto mix. It is whether `owl:sameAs`, used in instance data, mix.

dmitrizagidulin:
@OR13 @iherman 
> I think the statement of @msporny is not whether OWL and crypto mix. It is whether owl:sameAs, used in instance data, mix.

^ +1, thank you @iherman, that better describes what I meant. It's not that we can't use OWL to reason about security ontologies. It's that `owl:sameAs` implies "merge this particular instance graph with these key values into one graph", hence @msporny's `"asserting equivalence between any two direct or linked cryptographic identifiers is just asking for a world of pain security vulnerabilities. Introducing this would violate a key lemma in cryptography -- cryptographic identifiers are unique."`

OR13:
@dmitrizagidulin thats like saying that the `==` is dangerous.... don't use it, if you don't want to.... its not like there are not other vocabularies that can be used.... OWL is just one of the most popular standard ones.... if `owl:sameAs` is supported, does that prevent you from using something else, like `===` or `sameSubject` ?

If my use case does not block your use case, and you don't care to use my use case... why should you block it?

The arguments so far seem to be more like "linked data and cryptography are dangerous when used together"... which reduces to "cryptography is dangerous"...

msporny:
@OR13 wrote:
> If my use case does not block your use case, and you don't care to use my use case... why should you block it?

The concern here is that what you are suggesting will backfire horribly and take not only your use case down, but the rest of the DID community down with it... every DID method would be affected by the security vulnerability we are pointing out.

If there is a feature X that one doesn't care about, and it doesn't negatively impact the entire ecosystem, sure let that feature in. However, we are not talking about that in this PR... we are saying: "Feature X is going to negatively impact the entire ecosystem, it is dangerous, and because of that we need to tap the brakes hard and take stock of Feature X and all of the ramifications it could have across the entire DID ecosystem."

> The arguments so far seem to be more like "linked data and cryptography are dangerous when used together"

No, those aren't the arguments. You're missing the point. :)

The argument is: "When you enable "JohnA" and "JohnB" to become exactly the same person, and you merge all of JohnA's attributes and JohnB's attributes, that is, all you have now is "John"... it becomes impossible to know where John's attributes came from, including his cryptographic keys, and we're pretty sure that changes the security attack surface of DID Documents in ways that we haven't contemplated before. So, let's slow down and do the analysis and NOT let this thing in right now until we've done that. OR, we can do something that's far more narrow than owl:sameAs and addresses the use case that has been put forward (DID Migration).

OR13:
I think you may be missing the point... if `did:example:123` and `did:example:456` both decide to say they are `owl:sameAs` `@realDonaldTrump`... thats their right... there is nothing about using linked data formats that guarantees any kind of Truth....

If something like `owl:sameAs` is a security vulnerability... then all of RDF is security vulnerability... we can't say 1 class of statements in RDF and another are different, without admitting that the technology itself is a vulnerability... 

Before processing ANY information, you need to decide if you trust it... and act accordingly... in the case where you do trust that did:example:123 really is http://example.com/123... and you really do want to merge graphs where both are used as identifiers.... and you really want to use OWL... thanks to the decentralized nature of linked data... nobody can stop you... it feels like not explicitly addressing this is the real security vulnerability... 

its not like I'm inventing a new term / concept in RDF... I just saying... people are going to use the most popular vocabulary...

I'd be totally ok with a defintion like `owl:sameAs`... "you probably don't know what you are doing when you use this, be careful..." but I think its much clearer why thats not acceptable.... some people know what it is, and how to use it... 

The assumption here appears to be that people can't be trusted to use a dangerous concept correctly... isn't it our job to document it sufficiently so that they know how dangerous it is, and can make decisions safely? doesn't this apply to replay attack prevention, selection of crypto suites, and not publishing your private keys or telling people what your password is?

iherman:
> If my use case does not block your use case, and you don't care to use my use case... why should you block it?
>
> I'd be totally ok with a defintion like `owl:sameAs`... "you probably don't know what you are doing when you use this, be careful..." but I think its much clearer why thats not acceptable.... some people know what it is, and how to use it...

A DID document (when using JSON-lD) is a piece of RDF. Ie, it is used in an Open World, ie, any user has the possibility of adding an owl:sameAs statement to it. The issue is not, therefore, to block its usage: that is not possible. However, by adding owl:sameAs to the security context file the message the group would convey is that using owl:sameAs is _THE_ way of stating the identity of two subjects. That is where, I believe, the disagreement is.

My feeling is that the [PR for a sameAs property in DID Core](https://github.com/w3c/did-core/pull/348) submitted by @talltree covers the very specific identity use case (which has the added value of providing a definition for the pure JSON an CBOR versions, too). However, we cannot do as if owl:sameAs didn't exist. We could (probably should) add a note into the spec why its usage is dangerous, though possible, while moving ahead with the aforementioned PR.

Would that be an acceptable way forward?

OR13:
@iherman yes, my proposal is document explicitly dangerous terms that might be used with the security context.... and ass you noted... this is the w3id.org security context... anyone can fork it... and the problem will not go away... we should defend people by being explicit.

OR13:
Stumbled across this schema.org example:

```
{
  "@context": "http://schema.org",
  "@type": "VisualArtwork",
  "name": "La trahison des images",
  "alternateName": "The Treachery of Images",
  "image": "http://upload.wikimedia.org/wikipedia/en/b/b9/MagrittePipe.jpg",
  "description": "The painting shows a pipe. Below it, Magritte...",
  "creator": [
    {
      "@type": "Person",
      "name": "René Magritte",
      "sameAs": "https://www.freebase.com/m/06h88"
    }
  ],
  "width": [
    {
      "@type": "Distance",
      "name": "940 mm"
    }
  ],
  "height": [
    {
      "@type": "Distance",
      "name": "635 mm"
    }
  ],
  "artMedium": "oil",
  "artworkSurface": "canvas"
}
```

iherman:
Note that, per the schema.org JSON-LD context file:

https://schema.org/docs/jsonldcontext.json

the schema.org ‘sameAs’ is *not* the OWL relationship.  It is also interesting to see the definition of the property:

https://schema.org/sameAs

“URL of a reference Web page that unambiguously indicates the item's identity. E.g. the URL of the item's Wikipedia page, Wikidata entry, or official website.”

It is of a very restricted usage, thus.

Cheers,

Ivan

P.S. Personally, I don’t like the fact that they use the same term because it can be misleading, but that is the way it is. 

——
Ivan Herman

(Written on my iPad. Excuses for brevity and misspellings...)

OR13:
I agree, i'd rather not see `sameAs` if it's not doing to be OWL's version.

OR13:
Discussion continues in the DID WG... summarizing here:

`owl:sameAs` is bi-directional (symmtric) and transitive... if those are properties you want, use `owl:sameAs`.
`alsoKnownAs` is unidirectional (asymmetric) and not(?) transitive... if those are properties you want, use `alsoKnownAs`.. wherever it gets defined.

@msporny and other OWL haters ; ) think OWL is dead, and that we should discourage its use by avoiding relying on it... yet a note that `owl:Thing` is the base class of all verification methods....

I think we are aligning on an information model that allows for representations to use `owl` if they wish (end of the day, open world model means you can't stop people anyway).... but perhaps some of the conversation regarding keypair vs verification method is relevant here as well: https://or13.github.io/best-linked-data-suites-2020/#information-model

OR13:
I continue to advocate for including OWL in security vocab, and DOCUMENTING the risks associated with symmetric transitive closure... defining security vocabulary means explaining how security works... its our job in this repo to warn people, and the stronger the objection to using `owl:sameAs`... the stronger the recommendation should be to explain why thats a security issue.

iherman:
> @msporny and other OWL haters ; ) think OWL is dead, and that we should discourage its use by avoiding relying on it... yet I note that `owl:Thing` is the base class of all verification methods....

I don't know whether you included me in the "the other" but I would prefer if you didn't make such remarks.

OWL is extremely powerful and it is clearly of a huge value in many different applications in, say, healthcare applications just to take this example. But it _exactly_ because it is so powerful that it is of a good diligence to be very careful when using it. Just to take an example, I don't think my questions in https://github.com/w3c-ccg/security-vocab/pull/45#issuecomment-654917770 have been fully answered, and there may be many of those. There may be (many) others.

msporny:
Here's a paper from back in 2009, co-authored by Ivan, that explains the complexities of owl:sameAs: https://www.w3.org/2009/12/rdf-ws/papers/ws21

> @msporny and other OWL haters

OWL is the basis for a variety of our vocabularies... it's fine there. It's when we move it into DIDs, VCs, and cryptography that things start to get hairy.

OR13:
@iherman @msporny I was joking about OWL haters comment, but not about needing to document this clearly :)

I am very interested in the overlap between formal logic, reasoning systems and cryptographic proof systems. 

Consider the value of a ZKP for `owl:sameAs`.

that being said, I agree with you both, that with great power comes great responsibility.

OR13:
@iherman sorry for not responding to this directly.

> > `sameAs`: One DID is asserted to be equivalent as another by the DID controller. Importantly, both DIDs are available / resolvable, and either can be used interchangeably (so, for example, if they're being used for an identity-based access control system, authenticating as did A is equivalent to authenticating to the sameAs DID B). Bi-directional references are required for validation.
> 
> Thanks for the example. A question: does it mean that the statements of the DID documents become all interchangeable? I.e., let us say the two DID-s are `D_1` and `D_2`, with `Doc_1` and `Doc_2` as DID documents, respectively. `Doc_1` defines, say, a service via the `service` predicate; are we sure that this service is valid for `D_2`, too (because that is what `owl:sameAs` would mean).
> 

yes, I think so... if your intention was for the service to only work with the D_1,  you would not use `owl:sameAs`... You would only use it when you intended for predicates to work for both identifiers, or if you intended to use its presence or absence to construct refined data models, represented in RDF or other systems ...after all, `owl:sameAs` is just another statement made by a controller.... its interpretation is up to software systems... if one is really worried about dangerous RDF statements, one should perform some kind of input validation on did documents before processing them... just like checking for a script tag in text before saving it to a database.

> Another question, but from another point of view. `owl:sameAs` is, strictly on a fix URI. So if we say `D1 owl:sameAs D2`, that statement does not have _any_ bearing on the respective DID URLs. I.e., if `Doc_1` has a statement on, say, `D_1#123`, that does _**not**_ mean that `D_1#123 owl:sameAs D_2#123`. Is that o.k., or may that be a problem?

^ this is great news, it means that all the keys are not merged, which means that proofs are still tied to a specific URL.... i think this was @msporny @dlongley 's main concern.

iherman:
> 
> > Another question, but from another point of view. `owl:sameAs` is, strictly on a fix URI. So if we say `D1 owl:sameAs D2`, that statement does not have _any_ bearing on the respective DID URLs. I.e., if `Doc_1` has a statement on, say, `D_1#123`, that does _**not**_ mean that `D_1#123 owl:sameAs D_2#123`. Is that o.k., or may that be a problem?
> 
> ^ this is great news, it means that all the DID Document member objects remain separate (services and verification methods), which means that proofs are still tied to a specific URL.... i think this was @msporny @dlongley 's main concern.

That is fine. Let us not forget, however, that this is a fine point that is not necessarily realized by everyone. (Fundamentally, URL-s are, for better or for worse, completely opaque strings from RDF’s point of view.) By using `owl:sameAs` false expectations may be raised, and that is a potential problem, too.

msporny:
> ^ this is great news, it means that all the DID Document member objects remain separate (services and verification methods), which means that proofs are still tied to a specific URL.... i think this was @msporny @dlongley 's main concern.

Wait... no :) -- I don't think @iherman is saying what you think he's saying... again, it's this subtleties that create the great concern around the use of owl:sameAs.

```
<D1> authentication <D1#key1> .
<D2> authentication <D2#key2> .
<D1> owl:sameAs <D2> .
```

The above could result in effectively this graph:

```
<D_merged> authentication <D_merged#key1> .
<D_merged> authentication <D_merged#key2> .
```

Thus, a compromise of D1 results in a compromise of D2 (unless we write a TON of new spec text into what you have to do to prevent that from happening).

iherman:
@msporny correct. What I meant was that if one document says

```
<D1#key1> whateverterm whatevervalue
```

this **_will not_** mean

```
<D2#key1> whateverterm whatevervalue
```

in spite of 

```
<D1> owl:sameAs <D2>
```

As you say, there are many subtleties that we may not realize at first glance.

msporny:
> ```
> <D1#key1> whateverterm whatevervalue
> ```
> 
> this **_will not_** mean
> 
> ```
> <D2#key1> whateverterm whatevervalue
> 
> As you say, there are many subtleties that we may not realize at first glance.
> ```

Yes, exactly, and all of these subtitles are the issue with `owl:sameAs`.

OR13:
@msporny @iherman 

I interpreted it comment correctly, but maybe I didn't clearly communicate why I think thats valuable to define.... consider:

https://w3c-ccg.github.io/security-vocab/#Ed25519VerificationKey2018

parent class = `owl:Thing`

```
{
  "id": "did:example:123#z6MkkQBvgvqb6zGvS4cydworpUaRDzpszSFixq49ahbDeUTG",
  "type": "Ed25519VerificationKey2018",
  "controller": "did:example:123",
  "publicKeyBase58": "6wvt6gb9mSnTKZnGxNr1yP2RQRZ2aZ1NGp9DkRdCjFft"
}
```

`did:example:123#z6MkkQBvgvqb6zGvS4cydworpUaRDzpszSFixq49ahbDeUTG` `controller` `did:example:123`
`did:example:123` `sameAs` `did:example:456`

Seems good that:

`did:example:123#z6MkkQBvgvqb6zGvS4cydworpUaRDzpszSFixq49ahbDeUTG` `controller` `did:example:123`
`did:example:456#z6MkkQBvgvqb6zGvS4cydworpUaRDzpszSFixq49ahbDeUTG` `controller` `did:example:123`

remain distinguishable... don't you agree?

iherman:
> `did:example:123#z6MkkQBvgvqb6zGvS4cydworpUaRDzpszSFixq49ahbDeUTG` `controller` `did:example:123`
> `did:example:456#z6MkkQBvgvqb6zGvS4cydworpUaRDzpszSFixq49ahbDeUTG` `controller` `did:example:123`

Yes, those are distinguishable, ie, an OWL reasoner will not, conceptually, "merge" those two statement. (I leave this to you guys to decide whether that is good or bad:-)

msporny:
> Yes, those are distinguishable, ie, an OWL reasoner will not, conceptually, "merge" those two statement. (I leave this to you guys to decide whether that is good or bad:-)

Sure, but that's not the attack we're concerned about. :)

OR13:
> > Yes, those are distinguishable, ie, an OWL reasoner will not, conceptually, "merge" those two statement. (I leave this to you guys to decide whether that is good or bad:-)
> 
> Sure, but that's not the attack we're concerned about. :)

@msporny please provide exact triples of the attack you are concerned about.

msporny:
> @msporny please provide exact triples of the attack you are concerned about.

I did in https://github.com/w3c-ccg/security-vocab/pull/45#issuecomment-680089537

OR13:
@msporny `<D_merged>` does not exist, thats not an actual example....

`did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP` `sameAs` `did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX`
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#authenticationMethod> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#authenticationMethod> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .


### DID 1
```
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6LScqmY9kirLuY22G6CuqBjuMpoqtgWk7bahWjuxFw5xH6G> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/security#X25519KeyAgreementKey2019> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6LScqmY9kirLuY22G6CuqBjuMpoqtgWk7bahWjuxFw5xH6G> <https://w3id.org/security#controller> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6LScqmY9kirLuY22G6CuqBjuMpoqtgWk7bahWjuxFw5xH6G> <https://w3id.org/security#publicKeyBase58> "2AbNdSuzFSpGvsiSPBfnamcKzk9Q3WRRpY2EToHZEuKW" .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/security#Ed25519VerificationKey2018> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#controller> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#publicKeyBase58> "dbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1" .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#assertionMethod> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#authenticationMethod> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#capabilityDelegationMethod> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#capabilityInvocationMethod> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#keyAgreementMethod> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6LScqmY9kirLuY22G6CuqBjuMpoqtgWk7bahWjuxFw5xH6G> .
<did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> <https://w3id.org/security#publicKey> <did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP> .
```

### DID 2
```
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/security#EcdsaSecp256k1VerificationKey2019> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#controller> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#publicKeyBase58> "d5cW2R53NHTTkv7EQSYR8YxaKx7MVCcchjmK5EgCNXxo" .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#assertionMethod> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#authenticationMethod> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#capabilityDelegationMethod> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#capabilityInvocationMethod> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .
<did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> <https://w3id.org/security#publicKey> <did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .
```

msporny:
> @msporny <D_merged> does not exist, thats not an actual example....

Yes, it does. That's what `owl:sameAs` does in a reasoner, it's pure equivalence: 

( a1 , a2 ) ∈ IEXT(I(owl:sameAs)) | iff | [a1 = a2](https://www.w3.org/TR/2012/REC-owl2-rdf-based-semantics-20121211/#topic-semcond-eqdis-informativenotes)

[SameIndividual](https://www.w3.org/TR/2012/REC-owl2-direct-semantics-20121211/#Assertions)( a1 ... an ) | (aj)I = (ak)I for each 1 ≤ j ≤ n and each 1 ≤ k ≤ n

Also, note what I'm having to point to to even have this discussion... this is why I want to avoid owl:sameAs -- very few people have the sort of math background to even interpret those specs I'm linking to? Even I'm starting to hit my limit... I had to stare at the specs and remind myself of even how to interpret OWL statements:

https://www.w3.org/TR/2012/REC-owl2-direct-semantics-20121211/#Interpretations

How many people will be able to even read that section above? This is just really dangerous ground...

OR13:
is `a1` === `did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP` ?

because `a1` != `a2` for:

`a1` == `did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP` && `a2` == `did:key:zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX`

However, if a reasoner would logically accept the following triple after merging:

`<DID1> <https://w3id.org/security#authenticationMethod> <DID2#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX> .`

I still see no problem with this... in an open world, representations of information are not restricted in any manner... anyone can create such a  structure, and IF they desired to use OWL, leverage the behavior of OWL reasoners...

The argument against OWL is equivalent to arguing that integrals should be banned because students don't always compute them correctly.

I'm fine with that as long as we make it clear that w3id.org/security is of the opinion that consumers are incapable of comprehending OWL and using it correctly... however, relying on `owl:Thing` as a base class for `Ed25519Verification2018` contradicts this assertion....

Picking and choosing when to allow OWL will lead to more confusion not less.... better not to use it at all, or use it consistently.

Here are some proposals:


#### PROP 1

Retain OWL in w3id.org/security and define sameAs, and warn about this behavior / feature.

#### PROP 2

Remove OWL from w3id.org/security entirely, do not include sameAs.

#### PROP 3

Retain OWL in w3id.org/security and don't define sameAs, wait for someone to shoot themselves in the foot.

I take it you remain in favor of PROP 3.

I would be in favor of PROP 1 and then PROP 2 in that order, PROP 3 is unacceptable to me.

iherman:
@OR13 @msporny can someone give me a link to the current vocabulary we are talking about? @OR13, I have difficulties to parse your proposals, and I would like to see what we are talking about... Thx

OR13:
@iherman 

https://w3c-ccg.github.io/security-vocab/#Ed25519VerificationKey2018
https://w3c-ccg.github.io/security-vocab/#Digest
https://w3c-ccg.github.io/security-vocab/#Bls12381G1Key2020
https://w3c-ccg.github.io/security-vocab/#Key

iherman:
> @iherman
> 
> A few of the vocabulary terms in https://w3id.org/security that rely on `owl:Thing`
> 
> https://w3c-ccg.github.io/security-vocab/#Ed25519VerificationKey2018
> https://w3c-ccg.github.io/security-vocab/#Digest
> https://w3c-ccg.github.io/security-vocab/#Bls12381G1Key2020
> https://w3c-ccg.github.io/security-vocab/#Key
> 

Thanks. If I understand it right, the only reason you refer to OWL (apart from `owl:sameAs`) is that you refer to `owl:Thing`.  But that leads to a bunch of questions:

- Why do you need `owl:Thing` in the first place? What is the statement you want to express with this?
- Is `Ed25519VerificationKey2018` indeed a class in the OWL sense (as is suggested in the English text)? Ie, is it a collection of individuals? What are those individuals that are of type `Ed25519VerificationKey2018` (this question may simply reflect my limited knowledge)? If it _is_ a class, why don't you use `owl:Class` (ie, typing the stuff as a class)?
- Why do you use the OWL vocabulary for this in the first place? Why not use `rdfs:Resource` or, if needed, `rdfs:Class`? 

The third set of questions is probably the most important. While, when looking at JSON-LD from an RDF point of view, the RDF(S) vocabulary is an integral part of the environment, referring to OWL pulls in a sledgehammer. One should not do that unless it is necessary. I do not personally agree with the strong anti-OWL sentiment that is sometimes present among Linked Data people, but it is correct that one should try to stay simple, ie, use RDF(S) only, if possible.

Coming back to your question, I am firmly in PROP 2 above. If a specific vocabulary or applications needs to use complex ontologies and would, therefore, pull in OWL, that is perfectly all right, possible, and justified. But there is no reason to do that from the start by introducing it in the core vocabulary definition.

OR13:
@iherman thanks for your comments!

I did not make these choices, I am trying to be consistent with the choices made before I joined the WG, I would be fine with PROP 2.

msporny:
> I did not make these choices

The existence of owl:Thing is a copy-paste from over a decade ago... :P

I'm fine w/ it being removed, but not for the reason @OR13 is suggesting. It's perfectly fine to pick and choose which parts of the Linked Data ecosystem to use if it makes sense for your use case. In this case, I think having `owl:Thing` is fine, but also agree with @iherman that we should use simpler things if we can get away with it, and in this use case, we can.

> I would be fine with PROP 2.

+1, let's go with PROP 2.

OR13:
@msporny what is a more generic base class that we might use, that would not invite folks to keep trying to use OWL?

regarding the assertion that its fine to pick and choose, for your use case.... i totally agree, but this is shared security vocab, and you clearly are not ok with me picking `owl:sameAs`... and I am not ok with retaining other OWL stuff, especially if its because of a copy paste error from 10 years ago.

Picking and choosing vocabulary that is different than w3id.org/security/v2 will result in non interoperable linked data signatures... maybe those vocabularies would be better documented, more internally consistent, make use of existing ontologies better, instead of defining a new vocabulary.... But I'd rather share a well documented security vocab that I don't 100% love, then go off and create my own version based on my aesthetic preferences.... 

My hope is that others will feel the same about w3id.org/security eventually and that they will help us keep relying on it for interoperability, as all of the VC Data Model and DID Core Linked Data community already is :) ... 

I would say that encouraging picking and choosing in this case, might lead to competing  versions of w3id.org/security/v2 that would cripple the interoperability we have for DID Core and VC Data Model... I wrote tests to confirm this a while back here:

https://github.com/decentralized-identity/context/blob/master/use-security-context.js

I was able to get all the Universal Resolver DID Documents to include no undefined properties, by mutating them to use a DIF context which defined all their properties... but in the process, I broke compatibility with W3C DID Core and W3C VC Data Model.... 

If `https://w3id.org/security#assertionMethod` changes, the canonical representation of DID Documents and VC Proofs changes... in ways that would probably be breaking for everyone.

All this to say, I take maintenance of the security vocab very seriously, and I see it s a critical dependency for both W3C DID Core and W3C VC Data Model.

iherman:
`rdfs:Resource` is a general base class, essentially like `owl:Thing`. But if we have something defined as a class, then it seems to be enough that this is of type `rdfs:Class`.

msporny:
> `rdfs:Resource` is a general base class, essentially like `owl:Thing`. But if we have something defined as a class, then it seems to be enough that this is of type `rdfs:Class`.

What @iherman said.

> regarding the assertion that its fine to pick and choose, for your use case.... i totally agree, but this is shared security vocab, and you clearly are not ok with me picking owl:sameAs... and I am not ok with retaining other OWL stuff, especially if its because of a copy paste error from 10 years ago.

That's fine, I said let's do PROP 2 ([above](https://github.com/w3c-ccg/security-vocab/pull/45#issuecomment-682805840)), which has us removing `owl:sameAs`. :)

OR13:
@msporny @iherman awesome, thanks for your patience on this, I can revise this PR to remove owl and rely on `rdfs`.

OR13:
Adopting PROP 2, opened https://github.com/w3c-ccg/security-vocab/pull/59 closing this PR.

OR13:
@TallTed based on your comments here: https://github.com/w3c-ccg/traceability-vocab/issues/152 

You might be interested in this thread.

FWIW, we also discussed this issue in did core:

https://github.com/w3c/did-core/pull/348

I personally would have loved to see more support and direction for integrating OWL into JSON-LD / LD Security.

TallTed:
@OR13 -- Is there a particular piece of this megathread you think I should see?

I've read a chunk, and most of it appears to come down to a misunderstanding of what `owl:sameAs` means, which is simply that any two IRIs in an `owl:sameAs` relation identify the same referent, the same entity; in other words, `owl:sameAs` is a relation of _coreference_.

I don't know what need there was or is to "integrate OWL into JSON-LD / LD-Security".  Neither do I see any barrier to using these things together without such "integration", whatever that might mean in this context.

OR13:
@TallTed mostly just the background as it may relate to VCs that attempt to assert `coreference`.

You are right that you can 100% use owl:sameAs without it being included in sec-v2 / v3 or did core.

The main argument was that any presence of `owl:sameAs` was dangerous in DIDs / VCs... did core ended up using alsoKnownAs from activity streams instead:

https://www.w3.org/ns/did/#properties

Which was considered less dangerous.

PR #53, opened 2020-07-29
feat: add bbs bls signature and key definitions

tplooker:
Adds the definitions for the BBS+ signatures and associated keys documented [here](https://w3c-ccg.github.io/ldp-bbs2020/)

Requires merging of https://github.com/w3c-ccg/ldp-bbs2020/pull/34 and https://github.com/w3c-ccg/ld-cryptosuite-registry/pull/35

### DIFF
--- a/index.html
+++ b/index.html
@@ -678,6 +678,193 @@ and verified.
           </pre>
       </section>
 
+      <section id="BbsBlsSignature2020" about="https://w3id.org/security#BbsBlsSignature2020"
+      typeof="rdfs:Class">
+
+        <h3>BbsBlsSignature2020</h3>
+          <p>
+A Linked Data signature is used for digital signatures on RDF Datasets.
+The default canonicalization mechanism is specified in the RDF Dataset
+Normalization specification, which deterministically names all
+unnamed nodes. Importantly, a BbsBlsSignature digests each of the statements
+produced by the normalization process individually to enable selective disclosure. 
+The signature mechanism uses Blake2B as the digest for each statement and produces a single 
+output digital signature.
+          </p>
+          <dl>
+            <dt>Status</dt>
+            <dd property="vs:term_status">unstable</dd>
+            <dt>Parent Class</dt>
+            <dd>Signature</dd>
+            <dt>Expected properties</dt>
+            <dd>verificationMethod, proofValue</dd>
+            <dt>Signature Properties</dt>
+            <dd>
+              <dl>
+                <dt>Default Canonicalization Algorithm</dt>
+                <dd>
+<a rel="canonicalizationAlgorithm"
+  href="https://w3id.org/rdf#URDNA2015">
+https://w3id.org/rdf#URDNA2015</a>
+                </dd>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -678,6 +678,193 @@ <h3>JsonWebSignature2020</h3>
...
+The signature mechanism uses Blake2B as the digest for each statement and produces a single 
+output digital signature.
+          </p>
+          <dl>
+            <dt>Status</dt>
+            <dd property="vs:term_status">unstable</dd>
+            <dt>Parent Class</dt>
+            <dd>Signature</dd>
+            <dt>Expected properties</dt>
+            <dd>verificationMethod, proofValue</dd>
```
OR13, line 700: I don't think you use `proofValue`... do you?

index.html
```diff
@@ -678,6 +678,193 @@ <h3>JsonWebSignature2020</h3>
...
+The example below shows how a basic JSON-LD proof is expressed in a
+JSON-LD snippet. Note that the proof property is directly embedded in the
+object. The signature algorithm specifies how the proof can be generated
+and verified.
+          </p>
+          <pre class="example prettyprint language-jsonld">
+{
+  "@context": ["https://w3id.org/security/v1", "http://json-ld.org/contexts/person.jsonld"]
+  "@type": "Person",
+  "name": "Dave Longley",
```
OR13, line 734: nice

index.html
```diff
@@ -678,6 +678,193 @@ <h3>JsonWebSignature2020</h3>
...
+        </p>
+          <p>
+The example below shows how a basic JSON-LD proof is expressed in a
+JSON-LD snippet. Note that the proof property is directly embedded in the
+object. The signature algorithm specifies how the proof can be generated
+and verified.
+          </p>
+          <pre class="example prettyprint language-jsonld">
+{
+  "@context": ["https://w3id.org/security/v1", "http://json-ld.org/contexts/person.jsonld"]
```
OR13, line 732: is this example correct? I would expect to see your context here.

index.html
OR13: nvm, i see you use it.

OR13: hmm wait... signatureValue != proofValue?

OR13: i guess we always use proofValue, so this is correct.

index.html
tplooker: :) just following the pattern @dlongley let me know if you would rather not be the example!

index.html
tplooker: Was just following the pattern of [JsonWebSignature2020](https://github.com/w3c-ccg/security-vocab/blob/0197e99b7c0eafb26c33d38803017f13f9e7e75f/index.html#L666), should I amend both?

OR13: Im not sure... I feel like a registry should always have examples that are as accurate as possible, but we're obviously still in the middle of cleaning up security vocab.... I think for now, leave it off, and just work really hard to get it into sec context v3 so we can do a big pass updating all these examples.

OR13:
This is pr against unstable, merging, names may change in the future.

PR #59, opened 2020-08-28
Replace owl:Thing with rdfs:Class

OR13:
Remove OWL from security vocab.

### DIFF
--- a/index.html
+++ b/index.html
@@ -87,7 +87,6 @@
   </head>
   <body prefix="rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
     rdfs: http://www.w3.org/2000/01/rdf-schema#
-    owl: http://www.w3.org/2002/07/owl#
     vs: http://www.w3.org/2003/06/sw-vocab-status/ns#
     sec: https://w3id.org/security#">
 
@@ -139,7 +138,7 @@ until it is ratified as an official document via the World Wide Web Consortium.
           <dt>Status</dt>
           <dd property="vs:term_status">unstable</dd>
           <dt>Parent Class</dt>
-          <dd>owl:Thing</dd>
+          <dd>rdfs:Class</dd>
           <dt>Expected properties</dt>
           <dd>type, created, verificationMethod, proofPurpose, jws</dd>
         </dl>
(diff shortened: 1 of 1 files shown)


iherman:
What does "parent Class" mean, and why do we need that? Isn't it clearer if that is replaced by saying 'of type rdfs:Class'?

OR13:
happy to make additional changes here.. @dlongley @msporny thoughts?

msporny:
> What does "parent Class" mean, and why do we need that? Isn't it clearer if that is replaced by saying 'of type rdfs:Class'?

This is another one of those things that's over a decade old. I don't have a strong opinion on keeping "Parent Class"... honestly, I'm not sure it even matters to list it. "Of Type" would be a strange header when rendered. We could do "Subclass of" or "Type of"... again, no strong preference.

iherman:
> > What does "parent Class" mean, and why do we need that? Isn't it clearer if that is replaced by saying 'of type rdfs:Class'?
> 
> This is another one of those things that's over a decade old. I don't have a strong opinion on keeping "Parent Class"... honestly, I'm not sure it even matters to list it. "Of Type" would be a strange header when rendered. We could do "Subclass of" or "Type of"... again, no strong preference.

I do not have an opinion on how it should appear editorially. From an RDF point of view, if we define a vocabulary that includes a class, what I am looking for is a triple of the form 

```
EcdsaSecp256k1Signature2019 rdf:type rdfs:Class
```

See, among the many examples, the [web annotation vocabulary](https://www.w3.org/ns/oa.jsonld). The term "Parent Class" is unclear and not a usual terminology in RDF vocabulary definitions and I would prefer to drop this terminology altogether. Stating somewhere clearly, in the English text, that this is a class seems to be perfectly fine to me.

OR13:
Given that the vocabulary is already seperated into `Classes` and properties, I would also be in favor of removing "Parent Class" entirely.

index.html
```diff
@@ -343,7 +330,7 @@ <h3>EncryptedMessage</h3>
       </section>
 
       <section id="GraphSignature2012" about="https://w3id.org/security#GraphSignature2012"
-        typeof="rdfs:Class">
+        typeof="sec:Signature">
```
OR13, line 333: @msporny @dlongley @davidlehn @iherman is this kind of annotation actually useful?

OR13:
@msporny I made some changes since you reviewed.

OR13:
I will merge this in a few days without objection.

PR #64, opened 2020-10-14
Remove old Web Payments Security Vocabulary page

clehner:
I found the presence of this page confusing when preparing to make a contribution, as it contains duplicate material to `index.html` and I was not sure at first which to edit. It appears to be an old page last updated in 2016, a copy of https://web-payments.org/specs/source/vocabs/security.html (https://github.com/web-payments/web-payments.org/blob/master/specs/source/vocabs/security.html) which is a draft published by the [Web Payments CG](https://www.w3.org/community/webpayments/).

I propose to remove this page from this security-vocab repo, as it appears to be a vestigial file.  It can continue to exist for historical purposes in the above mentioned locations and in this repo's git history. I checked backlinks on Ahrefs.com and Majestic.com and found none linking to the page under this repo's GitHub Pages.

If there is a reason to keep the page, maybe an explanation of it could be added to the readme. Or there could just be a link to the draft from the Web Payments Community Group at its canonical URL.


OR13 (approved):
agreed, its confusing and should be removed.

msporny:
It's been so long that I don't remember why we had kept this in here. It's definitely from the old Web Payments CG days:

https://web-payments.org/specs/

... but that site has it's own copy of the security vocabulary, so I think we're good here.

Agreed, let's remove it.

PR #66, opened 2020-10-15
add MerkleProof2019 definition

lemoustachiste:
So I naively copied a section and made some educated as to what should go where, but it may be incomplete (hopefully it is correct though). 

Please let me know if anything needs changing.

tagging @kimdhamilton as was one of the authors of the spec.

### DIFF
--- a/index.html
+++ b/index.html
@@ -584,7 +584,7 @@ and verified.
   },
   "proof": {
     "type": "MerkleProof2019",
-    "created": "2020-11-03T14:13:42.808099",
+    "created": "2020-11-03T14:13:42.808099Z",
     "proofValue": "zMcm4LfQFUZkWZyLpVrzYzaHRssD3xGQ6Yh6nEhErjfAqnxDWo7gBawdhynnbiucoqqwye54oTs5p6frH385nVvBf2UXMgRVtBZLq6tHN1MY1hQBMBBtXbWbYzj99GotAJRs7PxcvQmUQKHs8hkokZhgw92CNAtvDn2jeycvuyFPtWorxZaGP23mDkgwpberKmAqcPcsUFNooGe1UkQj8wws8D9paKHwFWxyCMvXNuYUGaBdZDaSpcd7uv3kEwgZAzAHb1jMgDeLSN1EdZA33LhPj5ACxMKs3vmwpTnjpNiALoZpJan",
     "proofPurpose": "assertionMethod",
     "verificationMethod": "did:example:23adb1f712ebc6f1c276eba4dfa#key-1"
--- a/index.html
+++ b/index.html
@@ -543,8 +543,9 @@ uses a SHA-256 digest and ECDSA to perform the digital signature. More informati
               <dl>
                 <dt>Default Canonicalization Algorithm</dt>
                 <dd>
-<a rel="canonicalizationAlgorithm"
-  href="https://w3id.org/security#GCA2015">https://w3id.org/security#GCA2015</a>
+                    <a rel="canonicalizationAlgorithm"
+                       href="https://w3id.org/rdf#URDNA2015">
+                        https://w3id.org/rdf#URDNA2015</a>
                 </dd>
                 <dt>Default Signature Algorithm</dt>
                 <dd>
--- a/index.html
+++ b/index.html
@@ -566,7 +566,7 @@ and verified.
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
-    "https://www.blockcerts.org/schema/3.0-alpha/context.json",
+    "https://www.w3id.org/blockcerts/v3.0-alpha",
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
   "id": "urn:uuid:bbba8553-8ec1-445f-82c9-a57251dd731c",
@@ -584,10 +584,10 @@ and verified.
   },
   "proof": {
     "type": "MerkleProof2019",
-    "created": "2020-10-15T15:16:52.191052",
-    "proofValue": "zMcm4LfQFUZkWZyLpVrzYzaHRssD3xGQ6Yh6nEhErjfAqnxDWo7gBawdhynnbiucoqqwye54oTs5p6frH385nVvBf2UXMgRVtBZLq6tHN1MY1hQBMBBtXbWbYzj99GotAJRs7PxcvQmUQKHs8hkokZhgw92CNAtvDn2jeycvuyFPtWorxZaGP23mDkgwpberKmAqcPcsUFNooGe1UkQj8wws8D9pdpAVmuKf5Zo78iq1tUWSUdGwN9sK57SQP6YTtL59b7WUK1NEHD1GL8o5Z2mY6MnxuxfiqkpjepaPVqTPfgmaCuS",
+    "created": "2020-11-03T14:13:42.808099",
+    "proofValue": "zMcm4LfQFUZkWZyLpVrzYzaHRssD3xGQ6Yh6nEhErjfAqnxDWo7gBawdhynnbiucoqqwye54oTs5p6frH385nVvBf2UXMgRVtBZLq6tHN1MY1hQBMBBtXbWbYzj99GotAJRs7PxcvQmUQKHs8hkokZhgw92CNAtvDn2jeycvuyFPtWorxZaGP23mDkgwpberKmAqcPcsUFNooGe1UkQj8wws8D9paKHwFWxyCMvXNuYUGaBdZDaSpcd7uv3kEwgZAzAHb1jMgDeLSN1EdZA33LhPj5ACxMKs3vmwpTnjpNiALoZpJan",
     "proofPurpose": "assertionMethod",
-    "verificationMethod": "ecdsa-koblitz-pubkey:0x7e30a37763e6Ba1fFeDE1750bBeFB4c60b17a1B3"
+    "verificationMethod": "did:example:23adb1f712ebc6f1c276eba4dfa#key-1"
   }
 }
           </pre>
--- a/index.html
+++ b/index.html
@@ -550,8 +550,8 @@ uses a SHA-256 digest and ECDSA to perform the digital signature. More informati
                 <dt>Default Signature Algorithm</dt>
                 <dd>
 <a rel="signatureAlgorithm"
-  href="http://www.w3.org/2000/09/xmldsig#rsa-sha256">
-http://www.w3.org/2000/09/xmldsig#rsa-sha256</a>
+  href="https://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256">
+https://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256</a>
                 </dd>
               </dl>
             </dd>
--- a/index.html
+++ b/index.html
@@ -519,6 +519,79 @@ and verified.
           </pre>
       </section>
 
+
+      <section id="MerkleProof2019" about="https://w3id.org/security#MerkleProof2019">
+
+        <h3>MerkleProof2019</h3>
+          <p>
+A Linked Data signature is used for digital signatures on RDF Datasets.
+The default canonicalization mechanism is specified in the RDF Dataset
+Normalization specification, which effectively deterministically names all
+unnamed nodes. The default signature mechanism
+uses a SHA-256 digest and ECDSA to perform the digital signature. More information at
+              <a href="https://w3c-ccg.github.io/lds-merkle-proof-2019/">https://w3c-ccg.github.io/lds-merkle-proof-2019/</a>
+          </p>
+          <dl>
+            <dt>Status</dt>
+            <dd property="vs:term_status">stable</dd>
+            <dt>Parent Class</dt>
+            <dd>LinkedDataSignature</dd>
+            <dt>Expected properties</dt>
+            <dd>proofValue</dd>
+            <dt>Signature Properties</dt>
+            <dd>
+              <dl>
+                <dt>Default Canonicalization Algorithm</dt>
+                <dd>
+<a rel="canonicalizationAlgorithm"
+  href="https://w3id.org/security#GCA2015">https://w3id.org/security#GCA2015</a>
+                </dd>
+                <dt>Default Signature Algorithm</dt>
+                <dd>
+<a rel="signatureAlgorithm"
+  href="http://www.w3.org/2000/09/xmldsig#rsa-sha256">
+http://www.w3.org/2000/09/xmldsig#rsa-sha256</a>
+                </dd>
+              </dl>
+            </dd>
+          </dl>
+          <p>
+The example below shows how a basic JSON-LD signature is expressed in a
+JSON-LD snippet. Note that the proof property is directly embedded in the
+object. The signature algorithm specifies how the signature can be generated
+and verified.
+          </p>
+          <pre class="example prettyprint language-jsonld">
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://www.blockcerts.org/schema/3.0-alpha/context.json",
+    "https://www.w3.org/2018/credentials/examples/v1"
+  ],
+  "id": "urn:uuid:bbba8553-8ec1-445f-82c9-a57251dd731c",
+  "type": [
+    "VerifiableCredential",
+    "BlockcertsCredential"
+  ],
+  "issuer": "did:example:23adb1f712ebc6f1c276eba4dfa",
+  "issuanceDate": "2010-01-01T19:33:24Z",
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "alumniOf": {
+      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1"
+    }
+  },
+  "proof": {
+    "type": "MerkleProof2019",
+    "created": "2020-10-15T15:16:52.191052",
+    "proofValue": "zMcm4LfQFUZkWZyLpVrzYzaHRssD3xGQ6Yh6nEhErjfAqnxDWo7gBawdhynnbiucoqqwye54oTs5p6frH385nVvBf2UXMgRVtBZLq6tHN1MY1hQBMBBtXbWbYzj99GotAJRs7PxcvQmUQKHs8hkokZhgw92CNAtvDn2jeycvuyFPtWorxZaGP23mDkgwpberKmAqcPcsUFNooGe1UkQj8wws8D9pdpAVmuKf5Zo78iq1tUWSUdGwN9sK57SQP6YTtL59b7WUK1NEHD1GL8o5Z2mY6MnxuxfiqkpjepaPVqTPfgmaCuS",
+    "proofPurpose": "assertionMethod",
+    "verificationMethod": "ecdsa-koblitz-pubkey:0x7e30a37763e6Ba1fFeDE1750bBeFB4c60b17a1B3"
+  }
+}
+          </pre>
+      </section>
+
       <section id="Ed25519VerificationKey2018" about="https://w3id.org/security#Ed25519VerificationKey2018"
       typeof="rdfs:Class">
         <h3>Ed25519VerificationKey2018</h3>
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -75,6 +75,9 @@
       "blockchainAccountId": {
         "@id": "https://w3id.org/security#blockchainAccountId"
       },
+      "MerkleProof2019": {
+        "@id": "https://w3id.org/security#MerkleProof2019"
+      },
       "Bls12381G1Key2020": {
         "@id": "https://w3id.org/security#Bls12381G1Key2020"
       },


index.html
```diff
@@ -519,6 +519,79 @@ <h3>LinkedDataSignature2016</h3>
...
+
+      <section id="MerkleProof2019" about="https://w3id.org/security#MerkleProof2019">
+
+        <h3>MerkleProof2019</h3>
+          <p>
+A Linked Data signature is used for digital signatures on RDF Datasets.
+The default canonicalization mechanism is specified in the RDF Dataset
+Normalization specification, which effectively deterministically names all
+unnamed nodes. The default signature mechanism
+uses a SHA-256 digest and ECDSA to perform the digital signature. More information at
```
msporny, line 531: > The default signature mechanism uses a SHA-256 digest and ECDSA to perform the digital signature.

Not sure if this is true; I don't think it is. The proofValue encodes the anchoring point on the blockchain, right? You don't use SHA-256 or ECDSA, right?

index.html
```diff
@@ -519,6 +519,79 @@ <h3>LinkedDataSignature2016</h3>
...
+                <dt>Default Canonicalization Algorithm</dt>
+                <dd>
+<a rel="canonicalizationAlgorithm"
+  href="https://w3id.org/security#GCA2015">https://w3id.org/security#GCA2015</a>
+                </dd>
+                <dt>Default Signature Algorithm</dt>
+                <dd>
+<a rel="signatureAlgorithm"
+  href="http://www.w3.org/2000/09/xmldsig#rsa-sha256">
+http://www.w3.org/2000/09/xmldsig#rsa-sha256</a>
```
msporny, line 553: The signature algorithm is different here... you need to point to the signature algorithm in the MerkleProof2019 spec.

index.html
```diff
@@ -519,6 +519,79 @@ <h3>LinkedDataSignature2016</h3>
...
+The example below shows how a basic JSON-LD signature is expressed in a
+JSON-LD snippet. Note that the proof property is directly embedded in the
+object. The signature algorithm specifies how the signature can be generated
+and verified.
+          </p>
+          <pre class="example prettyprint language-jsonld">
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://www.blockcerts.org/schema/3.0-alpha/context.json",
```
msporny, line 568: You may want to move this to a more permanent domain than just Blockcerts. I suggest https://w3id.org/blockcerts/v1 and redirecting that to https://www.blockcerts.org/schema/3.0-alpha/context.json.

index.html
```diff
@@ -519,6 +519,79 @@ <h3>LinkedDataSignature2016</h3>
...
+    "alumniOf": {
+      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1"
+    }
+  },
+  "proof": {
+    "type": "MerkleProof2019",
+    "created": "2020-10-15T15:16:52.191052",
+    "proofValue": "zMcm4LfQFUZkWZyLpVrzYzaHRssD3xGQ6Yh6nEhErjfAqnxDWo7gBawdhynnbiucoqqwye54oTs5p6frH385nVvBf2UXMgRVtBZLq6tHN1MY1hQBMBBtXbWbYzj99GotAJRs7PxcvQmUQKHs8hkokZhgw92CNAtvDn2jeycvuyFPtWorxZaGP23mDkgwpberKmAqcPcsUFNooGe1UkQj8wws8D9pdpAVmuKf5Zo78iq1tUWSUdGwN9sK57SQP6YTtL59b7WUK1NEHD1GL8o5Z2mY6MnxuxfiqkpjepaPVqTPfgmaCuS",
+    "proofPurpose": "assertionMethod",
+    "verificationMethod": "ecdsa-koblitz-pubkey:0x7e30a37763e6Ba1fFeDE1750bBeFB4c60b17a1B3"
```
msporny, line 589: You may want to use a multibase encoded multikey value here... this needs to be, at a minimum, a URI -- don't think ecdsa-koblitz-pubkey is a URI? Can this value be expressed as a DID? That would be best.

msporny (changes requested):
Minor request for changes, overall, good direction.

index.html
OR13: `"verificationMethod": "ecdsa-koblitz-pubkey:0x7e30a37763e6Ba1fFeDE1750bBeFB4c60b17a1B3"` this is very poorly supported form 2016.... see

https://github.com/w3c-ccg/lds-jws2020

https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/

OR13: also thats an ethereum address not a public key.

OR13: See also: https://github.com/decentralized-identity/EcdsaSecp256k1RecoverySignature2020

index.html
OR13: example: https://github.com/perma-id/w3id.org/pull/1891

index.html
lemoustachiste: so I based this description on this introduction here: https://w3c-ccg.github.io/lds-merkle-proof-2019/#introduction. 

The content (of the certificate, and furthermore of the merkle tree) is hashed before being saved on the blockchain. While I have worked less on the implementation of the issuance, during verification, to compare the local certificate to the one that was saved on the blockchain (via the proof), we hash with sha-256 to retrieve the expected value: https://github.com/blockchain-certificates/cert-verifier-js/blob/master/src/inspectors/computeLocalHash.js#L108.

However now that I am reading the code again, I think I made a mistake on the canonicalization algorithm: https://github.com/blockchain-certificates/cert-verifier-js/blob/master/src/inspectors/computeLocalHash.js#L84

index.html
lemoustachiste: see point above? I think it answers the same thing.

index.html
lemoustachiste: yeah this is still WIP and not really available to the outside world. I used this document as an example (verifiable) v3 blockcerts because this is what I have access to at this moment. 

But point taken this is something that we need to do on the roadmap to deliver blockcerts v3. If you feel the example should be more sturdy I will update.

index.html
lemoustachiste: Yeah at this point this is a configuration in my local `[cert-issuer](https://github.com/blockchain-certificates/cert-issuer/)`, that I have probably taken from @antonyronning. Again this does not effect immediately the proofing mechanism as this is a user set parameter, but if you find that confusing, again I can try and amend the example, no problem.

lemoustachiste:
@msporny @OR13 @tplooker friendly reminder about this PR

index.html
OR13: If you don't use ECDSA in the proof you should remove the term.... probably saying something like "The default proof mechanism relies on merkleproofs"

index.html
OR13: unless you are signing with ECDSA or RSA as part of contructing a MerkleProof2019, it does not.

index.html
OR13: I would recommend setting up w3id.org sooner rather than later, it will help if things needs to move and you don't want to break your proofs.

index.html
OR13: I recommend making `verificationMethod` a URI or DID.

index.html
```diff
@@ -519,6 +519,80 @@ <h3>LinkedDataSignature2016</h3>
...
+            <dt>Expected properties</dt>
+            <dd>proofValue</dd>
+            <dt>Signature Properties</dt>
+            <dd>
+              <dl>
+                <dt>Default Canonicalization Algorithm</dt>
+                <dd>
+                    <a rel="canonicalizationAlgorithm"
+                       href="https://w3id.org/rdf#URDNA2015">
+                        https://w3id.org/rdf#URDNA2015</a>
```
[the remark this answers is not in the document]
OR13: URDNA2015 is an RDF dataset normalization algorithm, its used to canonicalize JSON-LD before it is signed (both the document and the proof meta data such as `created`, `type` and `verificationMethod`. See https://w3c-ccg.github.io/ld-proofs/#create-verify-hash-algorithm

Im not sure exactly how MerkleProof2019 does or does not use this parameter.

OR13 (changes requested):
If its possible to include less descriptive text and more links to how MerkleProof2019 works, that might help get this merged faster.... I share @msporny 's concerns that what is being said here may not reflect how MerkleProof2019 actually works, but I am not an expert.

index.html
OR13: I looked at the code, the merkle proof does not use any digital signatures, it does use sha-256.

dsemenovsky: Hey @lemoustachiste @msporny @OR13 ,

I recall basing **MerkleProof2019** definition off the 2017 one.

ECDSA part refers to the process of verification of the credential since it is verified by storing a value (MerkleRoot hash) in a Blockchain transaction's input.

ECDSA is the algorithm that is used to sign the transaction using a wallet and I believe it can remain in the specification (potentially as a default one).

In this case in order to verify that the transaction with _transactionHash_ stored in the _proofValue_ was indeed signed with a wallet that belongs to a certain _DID_ you would use ECDSA to verify the blockchain transaction signature.

Does the explanation make sense?

OR13:
```
"proofValue": "zMcm4LfQFUZkWZyLpVrzYzaHRssD3xGQ6Yh6nEhErjfAqnxDWo7gBawdhynnbiucoqqwye54oTs5p6frH385nVvBf2UXMgRVtBZLq6tHN1MY1hQBMBBtXbWbYzj99GotAJRs7PxcvQmUQKHs8hkokZhgw92CNAtvDn2jeycvuyFPtWorxZaGP23mDkgwpberKmAqcPcsUFNooGe1UkQj8wws8D9pdpAVmuKf5Zo78iq1tUWSUdGwN9sK57SQP6YTtL59b7WUK1NEHD1GL8o5Z2mY6MnxuxfiqkpjepaPVqTPfgmaCuS",
    "proofPurpose": "assertionMethod",
    "verificationMethod": "ecdsa-koblitz-pubkey:0x7e30a37763e6Ba1fFeDE1750bBeFB4c60b17a1B3"
```

how do I get the transaction from this so I can do ECRecover / ECDSA verify to confirm the credential is anchored?

I think the part I am getting tripped up on is that a "merkle proof" does not use anything but sha-256.... and verifying a blockchain transaction that anchored a merkle root is not necessary to verify a merkle proof.

I'm not a merkle proof expert, so maybe I am wrong about this...

aren't you hashing the credential to a digest, batching the credential digest together with many others, and issueing a receipt, so you can prove that:

credential digest + receipt => is anchored in merkle root.

the merkle root could be anywhere, and anyone who has the merkle root and the receipt can verify the credential.

I assume the proofValue is the receipt, but then how do I get to the merkle root?

OR13: I would expect `verificationMethod` to point to a blockchain transaction that anchored a merkle root, not a public key or ethereum address that could have been used to sign a transaction...

dsemenovsky: @OR13 this is a very good question.
I will try to make a more detailed job explaining how both are connected tomorrow.

In a nutshell a Merkle Tree is an algorithm by which you can create a set of hashes that will lead down to the *leaf* which is a SHA256 of a particular credential in an array of credentials (to allow batch issuance of the credentials and confirming many credentials with one blockchain transaction).

The *merkleRoot* of the MerkleTree is another SHA256 that is not necessarily equal to the *targetHash*.

Now, these values partially create *proofValue* which is a MultiBase hash of a CBOR encoded object.

The other part is the transactionHash of the blockchain transaction.

After adding the transactionHash to the proofValue object it gets encoded.

Again, this is very highlevel and I am happy to go into the details.

In the meantime, I’ve got an OpenSource implementation for this here:

https://github.com/vaultie/lds-merkle-proof-2019

Deciding the proofValue will give you the transactionHas you are looking for.

The verificationMethod serves as a pointer to which DID was used to create this proofValue for the credential, in other words to verify **who** exactly issued the credential, not only to prove that it was issued.

And this is where you need ECDSA for example. Because you can prove that this transaction was indeed signed by the said wallet which belongs to some person or organization.

OR13: if the transactionHash is in the proofValue and is checked against the verification method, that makes sense.

OR13 (approved):
questions answered in comments, appear to resolve issues, since this is just the registration, I think its better to keep the spec details in the spec and link, but what is added here is fine imo, it just raised a bunch of questions, which have since been answered.

index.html
dsemenovsky: This is up to the verifier to check that and I guess MerkleRoot2019 is specifically designed to do so.

Please, if you have any other questions, I’ll try to answer as soon as I can.

msporny:
> @msporny @OR13 @tplooker friendly reminder about this PR

I'm still concerned about this PR -- the responses don't address the concerns... verificationMethod is still this bespoke thing, for example. Trying to figure out how to have a high bandwidth discussion about this -- perhaps schedule some CCG time to discuss. @OR13 -- the issue here is that it's not just a registration, it's also a definition, right? 

Perhaps the way to deal with this is to put a big issue marker on the section stating that it's a work in progress and is expected to change.

dsemenovsky:
> > @msporny @OR13 @tplooker friendly reminder about this PR
> 
> I'm still concerned about this PR -- the responses don't address the concerns... verificationMethod is still this bespoke thing, for example. Trying to figure out how to have a high bandwidth discussion about this -- perhaps schedule some CCG time to discuss. @OR13 -- the issue here is that it's not just a registration, it's also a definition, right?
> 
> Perhaps the way to deal with this is to put a big issue marker on the section stating that it's a work in progress and is expected to change.

@msporny Please, keep me in the loop as we are using this method for production purposes.

index.html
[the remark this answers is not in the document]
msporny: @dsemenovsky -- thoughts here? Are you actually doing RDF Dataset Normalization to get the hash of the object and then anchoring it in the blockchain via a merkle tree hash?

index.html
msporny: @dsemenovsky -- the thing above states that you are using RSA to do the digital signature. I expect that you're not actually doing that. Correct?

index.html
msporny: +1, please take the 10 minutes it takes to do the PR for w3id.org -- it'll save you hours, to days, to weeks of pain in the future.

msporny: @lemoustachiste wrote:
> yeah this is still WIP and not really available to the outside world.

@dsemenovsky wrote:
> Please, keep me in the loop as we are using this method for production purposes.

Haha -- so, which is it? Is this not available to the outside world *and* it's being used in production?

index.html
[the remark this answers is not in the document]
dsemenovsky: Yes, I am doing that.

I will provide the link here tomorrow morning.

index.html
dsemenovsky: No, I am not using RSA anywhere.
What should serve as a signature algorithm here?

Should we define that as ECDSA because that how the blockchain transaction gets signed?

index.html
dsemenovsky: I guess the confusion is caused by the fact we represent two different organizations, one uses it already and one doesn’t.

index.html
msporny: Agreed. At present, there is no such URI scheme called ecdsa-koblitz-pubkey, making this value questionable as the expectation is that it should be a URI of some kind.

msporny:
> @msporny Please, keep me in the loop as we are using this method for production purposes.

Oof, yikes... that changes things, doesn't it. :)

Added some comments into the review... if you add an issue marker via:

```
<p class="issue">Put the warning about how this method is still under development and may change here</p>
```

... that will allow us to pull this PR in with the appropriate warning.

index.html
msporny: Yes, it should be defined as ECDSA... is there a URL that identifies ECDSA? I don't know for sure, but I'd look at the IANA registries first... JOSE Web Algorithms might give you a URL... don't know if there is a direct HTTP link to the exact ECDSA algorithm.

index.html
msporny: Ah, ok, I get it now. Erm, so since someone has shipped this in production, how terrible would it be if the URL for the context changed and the `verificationMethod` value changed? I'll assert that it's technically possible to do this and not affect your production deployment (you'd just update your library with a new code path for the corrected specification and retail the old code path when you see the old `@context` value).

In the future, please don't ship things into production w/o getting proper review on your spec. It puts a burden on the community to start solving issues in suboptimal ways (and takes volunteer effort away from other things for the fire drill).

dsemenovsky: For us changes in `verificationMethod` will not affect anything.

In terms of `@context` we are using `credentials/v1` in tandem with custom context defined, which I believe is the whole purpose of having an array of contexts there, so that you can have flexibility and scalability of the credentials form.

It is more related to verifying JSON-LD structure than the cryptographic verification.

index.html
dsemenovsky: We are using DID here, and it points directly to the key within that DID (as 1 DID can have many keys).

Since DID has a form of URN, can it be generalized to be a URN?

msporny: It's supposed to be a URL (you need to be able to retrieve the key information using the `verificationMethod` parameter and use that to verify the signature).

If you use a DID here, we're all good, except that the example is wrong :).

Can you propose a change to the example so it uses a DID instead of this weird `"ecdsa-koblitz-pubkey:0x7e30a37763e6Ba1fFeDE1750bBeFB4c60b17a1B3"` thing?

index.html
msporny: Yes, good.

Please change:

```
https://www.blockcerts.org/schema/3.0-alpha/context.json
```

to 

```
https://w3id.org/blockcerts/v3-alpha
```

and then setup the appropriate redirect at https://w3id.org/

OR13: here is an example of how to setup a redirect with w3id.org: https://github.com/perma-id/w3id.org/pull/1891

index.html
OR13: One possible solution would be to use did:key here... its functioanlly the same (but uses a public key instead of an ethereum address):

https://github.com/transmute-industries/did-key.js/blob/master/packages/test-vectors/src/did-core-conformance/secp256k1/secp256k1.json#L43

the verificationMethod would be:

`did:key:zQ3shU9mxTYfMTLceuppoNM5siowHvw3xBk21r855kwgNHNAX#zQ3shU9mxTYfMTLceuppoNM5siowHvw3xBk21r855kwgNHNAX`

dsemenovsky: Here is how we are using:
```
{
  ...
  verificationMethod: "did:vaultie:0x8b9ac093ce03ee542bd4a37d3df77ad33079419f#keys-1"}
```

index.html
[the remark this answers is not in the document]
dsemenovsky: @msporny Here is the link I promised:

https://github.com/vaultie/cert-issuer/blob/f8ab836065a23ef86eb72c38347f7575619d8917/src/proofs/addMerkleProof.js#L7

Using JSON-LD Normalization is essential to make sure the hashes are calculated in a uniform way.

index.html
msporny: Yes, that's good, please update example to reflect that.

index.html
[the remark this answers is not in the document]
msporny: Acknowledged, thank you! This line is good then, confirmed that you're using URDNA2015 for canonicalization.

index.html
msporny: Found it! https://tools.ietf.org/html/rfc6931#section-2.3.6

Looks like the URL you're looking for is:

http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256

index.html
```diff
@@ -519,6 +519,80 @@ <h3>LinkedDataSignature2016</h3>
...
+                </dd>
+                <dt>Default Signature Algorithm</dt>
+                <dd>
+<a rel="signatureAlgorithm"
+  href="http://www.w3.org/2000/09/xmldsig#rsa-sha256">
+http://www.w3.org/2000/09/xmldsig#rsa-sha256</a>
```
msporny, line 554:
```suggestion
  href="http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256">
http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256</a>
```

index.html
lemoustachiste: I will tend to that

lemoustachiste: https://w3id.org/blockcerts/v3.0-alpha is now a thing, I am in the process of updating the example

index.html
lemoustachiste: @msporny we had a quick discussion with @dsemenovsky and we are not sure what the actual issue is. 
SHA-256 is used as the canonicalization algorithm of the document, and ECDSA is the signature of the blockchain transaction. 

MerkleProof2019 is specifically designed to work with Blockchains since its anchoring is used as part of the CBOR encoded `proofValue`. You may see here an example of what is encoded into the `proofValue`. https://w3c-ccg.github.io/lds-merkle-proof-2019/#example-1

Now that also raises a question on my side, should we make mention of CBOR anywhere in this addition? If so, under what definition item should it be listed?

index.html
msporny: +1, wonderful!

index.html
```diff
@@ -519,6 +519,80 @@ <h3>LinkedDataSignature2016</h3>
...
+  "issuanceDate": "2010-01-01T19:33:24Z",
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "alumniOf": {
+      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1"
+    }
+  },
+  "proof": {
+    "type": "MerkleProof2019",
+    "created": "2020-11-03T14:13:42.808099",
```
msporny, line 587:
```suggestion
    "created": "2020-11-03T14:13:42.808099Z",
```

msporny (approved):
Looks stable enough to merge... thanks for sticking with the review @lemoustachiste and @dsemenovsky !

index.html
msporny: No need to mention CBOR anywhere in this addition since the MerkleProof2019 specification defines what it should be. So there is nothing more to do in this specific PR.

Thanks for the link to the example... some thoughts on MerkleProof2019 -- Example 1 is a little strange, in that the text encoding seems to be hex where the hashing algorithm is assumed to be SHA-256? I suggest that you encode the text example as a multibase encoded multihash value. This will make it so that the data self-identifies as a base-16 encoded sha-256 32 byte value (for example). That will make MerkleProof2019 completely forwards compatible if you ever have to change the base-encoding format (to base58/base64url, for example) or the hash format (to SHA-3, etc.). 

Just some food for thought.

msporny:
Normative, multiple reviews, changes requested and made, no objections, merging.

OR13:
I was first made aware of `@vocab` use to provide "default IRIs" by this PR.... Linking to the ongoing discussion regarding credentials/v2 https://github.com/w3c/vc-data-model/issues/753

PR #69, opened 2020-10-29
feat: add jsonld-lint check

tplooker:
Adds [jsonld-lint](https://github.com/mattrglobal/jsonld-lint) as a check for the .jsonld contexts


msporny:
+1 for this PR... please don't check in the package-lock -- I know some people love doing that, but it's unnecessary here. We'll have the CI/CD process do that.

msporny (changes requested):
+1 for this PR... please don't check in the package lock -- I know some people love doing that, but it's unnecessary here. We'll have the CI/CD process just npm install -- no need to rely on a package lock file.

tplooker:
@msporny ok have removed the lockfile it really doesn't matter in this case as they are all dev deps any way, had to update eslint though as it was tripping over the `use strict` declaration at the top of .js files and intepreting it as a violation of [this rule](https://eslint.org/docs/rules/no-loss-of-precision) for some reason.

msporny (approved):

Issue #51, opened 2020-07-20
Provide reusable context compatible with VC data model context


Fak3:
Would it be possible to cook up a context for the security vocabulary that is compatible with VC data model context? So that it would be simple to reuse terms from the security vocabulary in the verifiable claims and presentations.

Currently both contexts are conflicting on some protected terms - namely Ed25519Signature2018 and maybe others.

The following example fails with `jsonld.SyntaxError: Invalid JSON-LD syntax; tried to redefine a protected term.` :

```json
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/security/v1"
  ],
  "@type": ["VerifiableCredential"],
  "issuer": {
    "name": "My issuer"
  }
}
```
[open in playground](https://tinyurl.com/y2fcnhya)

dlongley:
Yes, this is possible -- it's just a matter of someone putting in the time to do it, including thinking about how to best accomplish it and testing it. For example, should we have a different context per signature/proof type -- so you just attach your signature/proof type context? Or should there be one comprehensive security/v3 context that includes all of the latest? Should we do both?

dlongley:
cc: @OR13 and @tplooker for opinions.

OR13:
yes, this seems like a critical thing to do.

kdenhartog:
Ran into this problem when trying to update the `BbsBlsSignature2020` to use the v3-unstable context. It's conflicting because it's bringing in the v2 context, which brings in the v1 context which has a few property collisions. My suggestion is that we should update the verifiable credential context to remove `EcdsaSecp256k1Signature2019`, `EcdsaSecp256r1Signature2019`, `Ed25519Signature2018`, `RsaSignature2018` and then VCs should use the VC context and the latest security context instead. 

```json
{
  "@context": {
    "@version": 1.1,
    "@protected": true,

    "id": "@id",
    "type": "@type",

    "VerifiableCredential": {
      "@id": "https://www.w3.org/2018/credentials#VerifiableCredential",
      "@context": {
        "@version": 1.1,
        "@protected": true,

        "id": "@id",
        "type": "@type",

        "cred": "https://www.w3.org/2018/credentials#",
        "sec": "https://w3id.org/security#",
        "xsd": "http://www.w3.org/2001/XMLSchema#",

        "credentialSchema": {
          "@id": "cred:credentialSchema",
          "@type": "@id",
          "@context": {
            "@version": 1.1,
[...]

PR #70, opened 2020-11-02
feat: updates v3-unstable context to follow best practices

kdenhartog:
This v3-unstable context update makes two major changes.
- redefines all of the v2 and v1 context terms in a json-ld 1.1 compliant way rather than inporting through the context.
- Makes the v3 context compliant with the VC v1 context so that there's not a protected term conflict.

### DIFF
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -3,8 +3,9 @@
     {
       "@version": 1.1
     },
-    "https://w3id.org/security/v2",
     {
+      "id": "@id",
+      "type": "@type", 
       "JsonWebKey2020": {
         "@id": "https://w3id.org/security#JsonWebKey2020"
       },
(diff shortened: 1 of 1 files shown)


kdenhartog:
Ok @dlongley thanks for all the advice through the various issues necessary to put this together. We got this working with the Bbs Signature Suite in a way that doesn't conflict with the V1 context. This should address #51 as well.

tplooker:
Whilst alphabetising the terms may be useful, it is making this review quite difficult, do you mind reverting and applying the changes to this context in a way that makes it easier to review?

msporny:
Overall, +1 for this PR, thank you for putting it together @kdenhartog. Here are my high-level concerns, none of which should block this from being merged (but we may want to take a week or two to try to address them):

1. This is positive change, and a significant change in direction to the way we've been doing the Security vocabularies. Ideally, we'd have one or more test suites paired to all of these changes. Have you run this context through a test framework that you have? If not, we need to make sure that at least the other libraries that use the v2 context continue to work with a v3 context -- getting feedback from implementers will take time.

2. I noticed that only some of the types use scoped contexts. We need all types to use scoped contexts, which is going to bloat the context, but I believe that's inevitable if we are going to ensure that this context can be easily mixed with other vocabularies. This also means tests for all the types.

3. There is some cruft in here that we may want to remove. For example, I don't believe anyone is using Equihash these days... we may want to drop that. Same for RsaSignature2018 -- if folks want to use that, they can use the v2 signature suite. So, we should aggressively cull things that no one is using.

4. I noticed some use of `{ "@id": "https://foo.example#term" }` that could be shortened to just `"https://foo.example#term"`.

That's it upon a cursory examination of the PR... would appreciate responses to the above.

contexts/security-v3-unstable.jsonld
```diff
@@ -1,170 +1,531 @@
...
   "@context": [
     {
-      "@version": 1.1
-    },
-    "https://w3id.org/security/v2",
-    {
-      "JsonWebKey2020": {
-        "@id": "https://w3id.org/security#JsonWebKey2020"
+      "@version": 1.1,
+      "id": "@id",
```
dlongley, line 5: Should also have `"@protected": true` at the top level of this context.

dlongley:
I don't have time to do a full review of the changes here right now, but they are heading in the right direction. Thanks!

OR13 (approved):

OR13 (changes requested):
@kdenhartog can you adddress manu's concerns or open issues to track addressing them, and fix the merge conflict.

I am in favor of merging this asap, and taking smaller PRs which are easier to review for specific issues after.

kdenhartog:
> @kdenhartog can you adddress manu's concerns or open issues to track addressing them, and fix the merge conflict.
> 
> I am in favor of merging this asap, and taking smaller PRs which are easier to review for specific issues after.

Been working on them today to address them all. Just keep getting sidetracked with other conversations as well. Looking to get this merged ASAP as well.

kdenhartog:
> Overall, +1 for this PR, thank you for putting it together @kdenhartog. Here are my high-level concerns, none of which should block this from being merged (but we may want to take a week or two to try to address them):
> 
>     1. This is positive change, and a significant change in direction to the way we've been doing the Security vocabularies. Ideally, we'd have one or more test suites paired to all of these changes. Have you run this context through a test framework that you have? If not, we need to make sure that at least the other libraries that use the v2 context continue to work with a v3 context -- getting feedback from implementers will take time.

So far, we've validated this using [BbsBlsSignatures2020 suite](https://github.com/mattrglobal/jsonld-signatures-bbs/commit/2673a0a077c232ca5be61b93339be547e5341635). As far as I can tell this remains backwards compatible and doesn't redefine terms in a breaking way (I've effectively just copied v1 and v2). In the case of BbsBlsSignature2020 suite the compact operation being performed by jsonld-signatures is doing so in a way that the type ends up being `sec:BbsBlsSignature2020` because of how jsonld-signatures is tied to the v2 context at the moment. Specifically in the suite we validate VCs used in some of the newly defined v3 terms, but does not attempt to validate every single property works as expected. I agree doing so would be beneficial as we go to a ld-proof WG, but don't quite have the bandwidth to take on that much work at this point.

>     2. I noticed that only some of the types use scoped contexts. We need all types to use scoped contexts, which is going to bloat the context, but I believe that's inevitable if we are going to ensure that this context can be easily mixed with other vocabularies. This also means tests for all the types.

Yup agreed tests for all the types would be good. Just don't have the bandwidth to handle that at this point.
 
>     3. There is some cruft in here that we may want to remove. For example, I don't believe anyone is using Equihash these days... we may want to drop that. Same for RsaSignature2018 -- if folks want to use that, they can use the v2 signature suite. So, we should aggressively cull things that no one is using.

Happy to do this before stabilizing, but want to make sure we agree to which terms are being dropped rather than me picking and choosing. I'm not aware of how all of the v1/v2 terms are being used all the time. Can we do this in a separate issue and I'll go through and clean them up after?

>     4. I noticed some use of `{ "@id": "https://foo.example#term" }` that could be shortened to just `"https://foo.example#term"`.

Should all be cleaned up now. Thanks.

>     5. There are now merge conflicts due to other PRs being recently merged.

Should be fixed now as well. Thanks.

> That's it upon a cursory examination of the PR... would appreciate responses to the above before we merge. Thank you again for all the hard work!

All fixed up now as well.

contexts/security-v3-unstable.jsonld
kdenhartog: Fixed as well now.

kdenhartog: The usage of `@protected` ended up breaking many different tests in the BBS suite. I haven't been able to dig in and figure out what is the cause of this yet. The error I'm getting is that I'm redefining protected terms though.

OR13: suggest opening an issue, and working to address it with better tests seperatly.

OR13 (approved):
This PR is getting large and has already spawned a number of smaller issues, I would like to merge it and address the issues raised seperatly.

peacekeeper:
> 1. Have you run this context through a test framework that you have?

We have some unit tests that use this context in our Java library [jsonld-common-java](https://github.com/decentralized-identity/jsonld-common-java).

This has a locally cached version of security-v3-unstable.jsonld [here](https://github.com/decentralized-identity/jsonld-common-java/tree/master/src/test/resources/foundation/identity/jsonld).

I just tried to simply replace the [current version](https://github.com/w3c-ccg/security-vocab/blob/master/contexts/security-v3-unstable.jsonld) with the [version proposed in this PR](https://github.com/w3c-ccg/security-vocab/blob/a82be1a72a32cb14985ac7bba51d9f7c87e655c9/contexts/security-v3-unstable.jsonld), and the unit test did **NOT** pass.

The error that is thrown is _"An attempt was made to redefine a protected term"_.

This may very well be a bug in our library or the underlying [titanium-json-ld](https://github.com/filip26/titanium-json-ld) rather than the context file. I'd be happy to dig deeper to find out, just wanted to quickly report this.

OR13:
@peacekeeper thanks for testing this!

I am still in favor to taking these changes, but perhaps we can add some CI checks to this repo next?

wonder if we could get some cross language JSON-LD tests going so we an know how changes like this impact , go, python, js and java via CI.

kdenhartog:
I've not had time to figure out the problem with this yet, but have narrowed the problem down to using the '@protected' term at the top level of the context. My plan is to attempt to define each term individually and see which term is causing the problem and go from there. I'd vote to hold off on merging until I can clean up that one issue first since it's definitely not working on multiple implementations as @peacekeeper shared.

OR13:
we should add something like what I'm working on here : https://github.com/w3c-ccg/traceability-vocab/pull/28

to this repo eventually.

kdenhartog:
Figured out the issue and playing around with a few ways to fix it at the moment.

What I've discovered is that the new terms (except the one's defined in the VC context) are encountering issues where the terms are being redefined within the V3 context itself.

For example, when using the `"@protected": true` at the top of the context document it's effectively adding the term to each property. So for nonce it's going from `"nonce": "https://w3id.org/security#nonce"`

to now being defined like so:

```
"nonce": { 
  "@id": "https://w3id.org/security#nonce", 
  "@protected": true  
},
```

So the problem is that the way that we originally defined terms like BbsBlsSignatureProof2020 means that we're defining terms in the property contexts and then redefining them at the top level context in a way that conflicts.

kdenhartog:
I've removed the `"@protected": true` term at the document so this should now be working @peacekeeper 

Additionally, I've opened issue #75 so we can decide the best way to update the context before stablizing. This way we can merge this PR and keep the fixes easier to review.

kdenhartog:
@msporny or @dlongley thoughts on merging with the additional created issues / response provided above?

contexts/security-v3-unstable.jsonld
```diff
@@ -167,7 +168,241 @@
...
+    "LinkedDataSignature2015": "https://w3id.org/security#LinkedDataSignature2015",
+    "LinkedDataSignature2016": "https://w3id.org/security#LinkedDataSignature2016",
+    "CryptographicKey": "https://w3id.org/security#Key",
+    "authenticationTag": "https://w3id.org/security#authenticationTag",
+    "canonicalizationAlgorithm": "https://w3id.org/security#canonicalizationAlgorithm",
+    "cipherAlgorithm": "https://w3id.org/security#cipherAlgorithm",
+    "cipherData": "https://w3id.org/security#cipherData",
+    "cipherKey": "https://w3id.org/security#cipherKey",
+    "created": {"@id": "http://purl.org/dc/terms/created", "@type": "http://www.w3.org/2001/XMLSchema#dateTime"},
+    "creator": "http://purl.org/dc/terms/createdcreator",
```
peacekeeper, line 223:
```suggestion
    "creator": {"@id": "http://purl.org/dc/terms/creator", "@type": "@id"},
```

peacekeeper:
> I've removed the `"@protected": true` term at the document so this should now be working @peacekeeper

Thanks @kdenhartog, I can confirm that this works for me now, and I agree with your analysis (I came to the same conclusion). I wonder, was this working for others, i.e. having `@protected` at the top level in https://github.com/w3c-ccg/security-vocab/commit/a82be1a72a32cb14985ac7bba51d9f7c87e655c9?

peacekeeper:
One additional change I am noticing in my unit tests is that now `"sec"` and `"xsd"` namespaces are no longer defined and used at the top level, is that intentional? Anyway, I'd also be fine with merging now and addressing this later.

dlongley:
We should have `@protected` at the top level -- and if that causes errors we should figure out what they are and why. A security context should not allow terms to be redefined without conforming to the protection rules.

contexts/security-v3-unstable.jsonld
dlongley: @kdenhartog -- it sounds like adding `@protected` may have successfully surfaced errors that weren't getting detected previously in the BBS suite. The cause of those errors should be determined, we shouldn't turn off protection just to avoid them.

kdenhartog:
> > I've removed the `"@protected": true` term at the document so this should now be working @peacekeeper
> 
> Thanks @kdenhartog, I can confirm that this works for me now, and I agree with your analysis (I came to the same conclusion). I wonder, was this working for others, i.e. having `@protected` at the top level in [a82be1a](https://github.com/w3c-ccg/security-vocab/commit/a82be1a72a32cb14985ac7bba51d9f7c87e655c9)?

They were working fine because the v1/v2 context weren't protecting the terms there, so the scoped context terms that were redefining them overrode the terms discovered in v1/v2 security vocab. Now that all terms are being brought into a single context without reference it's causing issues once the `@protected` term is added at the top level. This is because the scoped context terms aren't actually scoped from what I can tell.

>One additional change I am noticing in my unit tests is that now "sec" and "xsd" namespaces are no longer defined and used at the top level, is that intentional? Anyway, I'd also be fine with merging now and addressing this later.

Yeah see here for the discussion about those considerations: https://github.com/w3c-ccg/security-vocab/issues/73

> We should have @protected at the top level -- and if that causes errors we should figure out what they are and why. A security context should not allow terms to be redefined without conforming to the protection rules.

I've diagnosed them in the comments above. I agree that the terms should not be allowed to be redefined and am working on a PR to reintroduce the usage of `@protected` correctly in a way that won't cause these issues. However, I'm looking to get this PR through first and then provide focused updates through subsequent PRs given the size of the PR already. Since it's labeled as unstable, it shouldn't be in use in production systems yet anyways so incremental changes should be fine in my opinion.

contexts/security-v3-unstable.jsonld
kdenhartog: Thanks will fix this.

peacekeeper:
> I'm looking to get this PR through first and then provide focused updates through subsequent PRs given the size of the PR already.

Fine with me, this is definitely good work! But let's try to not leave it in a broken state for long, even if it's marked as unstable..

dlongley:
@kdenhartog,

>  I agree that the terms should not be allowed to be redefined and am working on a PR to reintroduce the usage of @protected correctly in a way that won't cause these issues. However, I'm looking to get this PR through first and then provide focused updates through subsequent PRs given the size of the PR already. Since it's labeled as unstable, it shouldn't be in use in production systems yet anyways so incremental changes should be fine in my opinion.

Ok, sounds good.

kdenhartog:
> Fine with me, this is definitely good work! But let's try to not leave it in a broken state for long, even if it's marked as unstable..

Yup, agreed. I've got a few proposals on how to handle this and we can consider the options and discuss in issue #75

kdenhartog:
Merging after multiple Oks. Further work on this will continue as follow up PRs to add the `@protected` back into this context.

PR #77, opened 2020-11-12
Add "sec" and "xsd" namespaces to RsaSignature2018

peacekeeper:
Those are missing (and possibly in other places too).


OR13 (approved):

tplooker:
I thought the resolution was to not use prefixes in this new context?

OR13:
@tplooker I think it is, I interpreted this as the context was broken.... / not valid.... I'd rather fix these namespace issues all at once.

Issue #71, opened 2020-11-05
Write test suite for v3 context


kdenhartog:
Follow up issue related to #70 to write a test suite to address point 1 and 2 from Manu.

kdenhartog:
Basically the same thing as what's been done over in the tracibility-vocab repo: https://github.com/w3c-ccg/traceability-vocab/pull/28

PR #78, opened 2020-11-19
feat: adds protected term from jsonld 1.1 to v3-unstable context

### LINKED ISSUE #75: Establish best practices for using protected term
> Figured out the issue and playing around with a few ways to fix it at the moment.
> 
> What I've discovered is that the new terms (except the one's defined in the VC context) are encountering issues where the terms are being redefined within the V3 context itself.
> 
> For example, when using the `"@protected": true` at the top of the context document it's effectively adding the term to each property. So for nonce it's going from `"nonce": "https://w3id.org/security#nonce"`
> 
> to now being defined like so:
> 
> ```
> "nonce": { 
>   "@id": "https://w3id.org/security#nonce", 
>   "@protected": true  
> },
> ```
> 
> So the problem is that the way that we originally defined terms like `BbsBlsSignatureProof2020` means that we're defining terms in the property contexts and then redefining them at the top level context in a way that conflicts.

Related to issues discovered during this PR: https://github.com/w3c-ccg/security-vocab/pull/70/#issuecomment-724385975

kdenhartog:
This PR implements options 2 from [this comment](https://github.com/w3c-ccg/security-vocab/issues/75#issuecomment-729836345) so we can close #75

It should be noted that v3-context at the moment is incompatible with the VC context because of the bug that @peacekeeper discovered. I've started work to define a 2020 VC context [here](https://github.com/mattrglobal/jsonld-signatures-bbs/blob/46f3b8ae0aa49ffce789411d9132d6768cb178ff/__tests__/__fixtures__/contexts/credential_vocab.json) as well which would remove the LDS suites from the VC context.

This is a working example that removes the LDS suites from the VC context and puts it under the 2020 URL. However there's still changes that need to be made to it to point the URLs at the VC spec definition of the properties and to update vc-js (take a look at the [whole PR](https://github.com/mattrglobal/jsonld-signatures-bbs/pull/89) in jsonld-signatures-bbs to see the direction I'm looking to go) so that it checks for either the 2018 VC context or 2020 VC context in the first position of the context array.

### DIFF
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -233,21 +233,40 @@
         "jws": "sec:jws",
         "nonce": "sec:nonce",
         "proofPurpose": {
-          "@id": "sec:proofPurpose",
-          "@type": "@vocab",
-          "@context": {
-            "@version": 1.1,
-            "@protected": true,
-
-            "id": "@id",
-            "type": "@type",
-
-            "sec": "https://w3id.org/security#",
-
-            "assertionMethod": {"@id": "sec:assertionMethod", "@type": "@id", "@container": "@set"},
-            "authentication": {"@id": "sec:authenticationMethod", "@type": "@id", "@container": "@set"}
-          }
-        },
+            "@id": "https://w3id.org/security#proofPurpose",
+            "@type": "@vocab",
+            "@context": {
+              "@version": 1.1,
+              "@protected": true,
+              "id": "@id",
+              "type": "@type",
+              "assertionMethod": {
+                "@id": "https://w3id.org/security#assertionMethod",
+                "@type": "@id",
+                "@container": "@set"
+              },
+              "authentication": {
+                "@id": "https://w3id.org/security#authenticationMethod",
+                "@type": "@id",
+                "@container": "@set"
+              },
+              "capabilityInvocation": {
+                "@id": "https://w3id.org/security#capabilityInvocationMethod",
+                "@type": "@id",
+                "@container": "@set"
+              },
+              "capabilityDelegation": {
+                "@id": "https://w3id.org/security#capabilityDelegationMethod",
+                "@type": "@id",
+                "@container": "@set"
+              },
+              "keyAgreement": {
+                "@id": "https://w3id.org/security#keyAgreementMethod",
+                "@type": "@id",
+                "@container": "@set"
+              }
+            }
+          },
         "proofValue": "sec:proofValue",
         "verificationMethod": {"@id": "sec:verificationMethod", "@type": "@id"}
       }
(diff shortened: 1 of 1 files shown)


peacekeeper (changes requested):
I tested this and I get the same error as before (_"An attempt was made to redefine a protected term"_).

I tracked it down and it's again because `proofPurpose` is defined inconsistently in 8 different places (e.g. under `RsaSignature2018` it's different than under `proof`).

peacekeeper:
> This PR implements options 2

How does this implement option 2? I thought option 2 is to define it once in a way that can be shared across all suites, rather than defining it separately for each suite?

kdenhartog:
> I tested this and I get the same error as before (_"An attempt was made to redefine a protected term"_).
> 
> I tracked it down and it's again because `proofPurpose` is defined inconsistently in 8 different places (e.g. under `RsaSignature2018` it's different than under `proof`).

Ahh I think you're right. This is still only half the solution that should work. The full one needs to properly redefine all the suites that overlap with the VC context in the new way. I'll push an update commit to fix this.

> > This PR implements options 2
> 
> How does this implement option 2? I thought option 2 is to define it once in a way that can be shared across all suites, rather than defining it separately for each suite?

I don't believe defining it once was an option at all. Option 2 was `Define proofPurpose a specific way for the security-v3 context and all suites that use it must use that definition.`

peacekeeper (approved):
Works great now, thanks!

msporny (approved):
This is great, @kdenhartog, thank you for all the hard work. This is probably our best example of a modern JSON-LD context now (don't use prefixes, use scoped/protected terms, etc.)... It looks good enough to me to merge, at a cursory glance level.

There are definitely a number of terms in here that we should remove, like the equihash stuff, the legacy stuff that JsonWebKey2020 cryptosuite  now covers, etc... but that can happen in a separate PR.

OR13 (approved):

peacekeeper:
@kdenhartog while `proofPurpose` seems to work fine now, there still seems to be a problem with `proofValue`, which is defined differently under `Ed25519Signature2020` than it is on the top level.

This again triggers the _"An attempt was made to redefine a protected term"_ error for me.

But I'd be fine with fixing this in a separate PR after merging this one.

kdenhartog:
> @kdenhartog while `proofPurpose` seems to work fine now, there still seems to be a problem with `proofValue`, which is defined differently under `Ed25519Signature2020` than it is on the top level.
> 
> This again triggers the _"An attempt was made to redefine a protected term"_ error for me.
> 
> But I'd be fine with fixing this in a separate PR after merging this one.

Interesting I hadn't noticed this issue. I'll take a look at it while we're awaiting @dlongley to get a review and try to get it fixed in this PR. If not, I'll file an issue and make sure it gets addressed in a separate PR.

Additionally, the tests for issue #71 will really help to adopt this. I'll work with @OR13 on this to make sure that we can identify all of these issues in general in subsequent PRs as well so that we can make it easier to identify these things without having to manually test each one everytime.

contexts/security-v3-unstable.jsonld
```diff
@@ -1,11 +1,11 @@
 {
-  "@context": [
-    {
+  "@context": [{
       "@version": 1.1
     },
```
dlongley, line 4:
```suggestion
  "@context": [{
      "@version": 1.1
    },
```

I don't think we need this `@version` stuff, let's remove it.

contexts/security-v3-unstable.jsonld
```diff
@@ -88,85 +103,113 @@
...
+                "@container": "@set"
               }
-            },
-            "proofValue": "https://w3id.org/security#proofValue",
-            "verificationMethod": {
-              "@id": "https://w3id.org/security#verificationMethod",
-              "@type": "@id"
             }
+          },
+          "proofValue": "https://w3id.org/security#proofValue",
```
dlongley, line 152: Should this have ` "@type": "https://w3id.org/security#multibase"` or not? Not sure what `BbsBlsSignature2020` uses for `proofValue`. But `Ed25519Signature2020` should have `proofValue` defined that way. This implies not defining it at the top level so that each suite can define it separately.

tplooker: @dlongley interesting will review but on first though multi-base sounds like a good idea.

contexts/security-v3-unstable.jsonld
tplooker: @dlongley alternatively, does a top level declaration of the jsonld processor version mean we dont have to repeat it at the scoped context level?

tplooker (approved):

contexts/security-v3-unstable.jsonld
dlongley: Term definitions in JSON-LD 1.1 are automagically upgraded to version 1.1 when 1.1 features are used -- and otherwise there's no difference in how they are processed from 1.0. This means that declaring a version is wholly unnecessary, but this wasn't figured out in the JSON-LD WG until just before 1.1 became REC -- which was *after* the VC context v1 was published.

In short, we really don't need the declaration at all. The only reason we *may* need it, is for backwards compatibility with the VC context, which was written *before* the need to use it was removed from JSON-LD 1.1. Even then, we should be able to update processors to ignore `@version` tags when comparing protected terms for equality (this would be the only reason for needing it wrt backwards compat w/VC context v1).

Also, regarding the use of scoped contexts, it's always important to remember that only what's *in scope* (what's in the scoped context) is guaranteed to be defined. So, when you use a type-scoped-context in a JSON-LD document somewhere, you need to remember that anything that isn't in its scoped-context may no longer be defined, having been removed by something further up the JSON tree. So we want to ensure we define anything that may be used within that scope.

contexts/security-v3-unstable.jsonld
kdenhartog: Currently it's not a multibase formatted value, but I'm in favor of switching to that. Right now we just use a base64 (not base64url) encoded value.

For now I've removed the top level definition so that we can define these differently which should fix the problem @peacekeeper discovered. I think this is the appropriate way to handle this because the older suites do not use multibase either which would create conflict between the older definitions and the newer ones unless we property scope them.

If anyone would prefer to see us do it differently or has some other considerations that should be addressed please open an issue and we can decide there.

contexts/security-v3-unstable.jsonld
kdenhartog: Sweet, removed all the usages of this term. Thanks for the insight too.

OR13:
12 days old, no objections, changes are to unstable, merging/

PR #80, opened 2020-12-01
Add ci tests for jsonld-sigantures and vc-js

OR13:
I fixed the protected term conflicts that were related to `proofPurpose`.

See https://github.com/mattrglobal/jsonld-signatures-bbs/issues/93


OR13:
github actions are down, will need to restart later.

OR13:
@kdenhartog I added these test to try and help with the problems you noted with protected terms and contexts... you can see some of your issues are not reproduced in CI on every merge :)

kdenhartog:
This looks amazing! I don't quite have enough time today to go through it in depth, but I'll be sure to get around to it by the end of this week when I get a bit of down time. Thanks for taking on this chunk of work @OR13

OR13:
Tests for Ed25519Signature2018, JsonWebSignature2020 and BbsBlsSignature2020 are now passing for `https://www.w3.org/2018/credentials/v2`

OR13:
Tests for Ed25519Signature2018, JsonWebSignature2020 and BbsBlsSignature2020 are now passing for https://w3id.org/security/v3-unstable

contexts/security-v3-unstable.jsonld
```diff
@@ -2,12 +2,53 @@
   "@context": [{
     "id": "@id",
     "type": "@type",
-    "@protected": true,
```
kdenhartog, line 5: Why are we dropping the protected value here?

contexts/security-v3-unstable.jsonld
```diff
@@ -2,12 +2,53 @@
...
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "domain": "https://w3id.org/security#domain",
+        "expires": {
+          "@id": "https://w3id.org/security#expiration",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "jws": "https://w3id.org/security#jws",
+        "nonce": "https://w3id.org/security#nonce",
+        "proofPurpose": {
```
kdenhartog, line 27: This is going to need to match the definition below or this will cause a redefinition conflict

js/__fixtures__/contexts/credentials-v2-unstable.json
```diff
@@ -0,0 +1,289 @@
...
+          "@container": "@graph"
+        },
+        "verifiableCredential": {
+          "@id": "cred:verifiableCredential",
+          "@type": "@id",
+          "@container": "@graph"
+        }
+      }
+    },
+    "JsonWebSignature2020": {
```
kdenhartog, line 115: We shouldn't be defining suites in the vc context in my opinion. Rather this should link to the external security v3 context. Otherwise we're going to always have to keep these two in sync which is much harder with the VC context needing more bureaucratic paperwork to update compared to the security-vocab.

js/__fixtures__/contexts/did-v1.json
```diff
@@ -0,0 +1,87 @@
...
+    "@context": {
+      "@version": 1.1,
+      "id": "@id",
+      "type": "@type",
+      "dc": "http://purl.org/dc/terms/",
+      "schema": "http://schema.org/",
+      "sec": "https://w3id.org/security#",
+      "didv": "https://w3id.org/did#",
+      "xsd": "http://www.w3.org/2001/XMLSchema#",
+      "EcdsaSecp256k1VerificationKey2019": "sec:EcdsaSecp256k1VerificationKey2019",
```
kdenhartog, line 11: This is going to cause a redefinition error once the protected is fixed. These should be dropped from here and a reference to the v3 context should be included as well here in my opinion.

js/__fixtures__/contexts/example-v1.json
```diff
@@ -0,0 +1,47 @@
+{
+    "@context": [{
+      "@version": 1.1
+    },"https://www.w3.org/ns/odrl.jsonld", {
+      "ex": "https://example.org/examples#",
+      "schema": "http://schema.org/",
+      "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+  
+      "3rdPartyCorrelation": "ex:3rdPartyCorrelation",
+      "AllVerifiers": "ex:AllVerifiers",
```
kdenhartog, line 10: Is this just a copy of the VC examples context?  I'd prefer not encourage short-form IRIs not that it's not considered best practice, should we change these to long form syntax and then update in the old examples repo as well?

js/__fixtures__/contexts/odrl.json
```diff
@@ -0,0 +1,200 @@
+{
```
kdenhartog, line 1: Same could be said for this one as what was left in example-v1.json

js/__fixtures__/credentials/bbs/credentialTemplate.json
```diff
@@ -0,0 +1,28 @@
+{
+    "@context": [
+      "https://www.w3.org/2018/credentials/v2",
```
kdenhartog, line 3: I believe we're going to make this `https://www.w3.org/2021/credentials/v1`

js/__fixtures__/credentials/bbs/deriveProofFrame.json
```diff
@@ -0,0 +1,14 @@
+{
+    "@context": [
+      "https://www.w3.org/2018/credentials/v2",
```
kdenhartog, line 3: same here and probably many others too I bet, gonna just say find and replace these and not mention the other places.

js/__fixtures__/credentials/bbs/verifiableCredential.json
```diff
@@ -0,0 +1,43 @@
...
+  "issuer": {
+    "id": "did:key:z5TcESXuYUE9aZWYwSdrUEGK1HNQFHyTt4aVpaCTVZcDXQmUheFwfNZmRksaAbBneNm5KyE52SdJeRCN1g6PJmF31GsHWwFiqUDujvasK3wTiDr3vvkYwEJHt7H5RGEKYEp1ErtQtcEBgsgY2DA9JZkHj1J9HZ8MRDTguAhoFtR4aTBQhgnkP4SwVbxDYMEZoF2TMYn3s"
+  },
+  "identifier": "83627465",
+  "name": "Permanent Resident Card",
+  "description": "Government of Example Permanent Resident Card.",
+  "issuanceDate": "2019-12-03T12:19:52Z",
+  "expirationDate": "2029-12-03T12:19:52Z",
+  "credentialSubject": {
+    "id": "did:key:z5TcESXuYUE9aZWYwSdrUEGK1HNQFHyTt4aVpaCTVZcDXQmUheFwfNZmRksaAbBneNm5KyE52SdJeRCN1g6PJmF31GsHWwFiqUDujvasK3wTiDr3vvkYwEJHt7H5RGEKYEp1ErtQtcEBgsgY2DA9JZkHj1J9HZ8MRDTguAhoFtR4aTBQhgnkP4SwVbxDYMEZoF2TMYn3s#zUC7LTa4hWtaE9YKyDsMVGiRNqPMN3s4rjBdB3MFi6PcVWReNfR72y3oGW2NhNcaKNVhMobh7aHp8oZB3qdJCs7RebM2xsodrSm8MmePbN25NTGcpjkJMwKbcWfYDX7eHCJjPGM",
```
kdenhartog, line 20: Would you be alright if we consistently use either `did:example` style dids (with mocking in the docLoader) or did keys throughout these? I'm happy with either just care to see that we keep it consistent. Which would you prefer?

js/__tests__/bbs/credentials-v2.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+const didKeyFixture = require('../../__fixtures__/did-key-bls12381.json');
+
+const k0 = didKeyFixture[0].g2['application/did+ld+json'];
+k0.id = k0.controller + k0.id;
+
+const credentialTemplate = require('../../__fixtures__/credentials/bbs/credentialTemplate.json');
+let verifiableCredential;
+
+const documentLoader = url => {
```
kdenhartog, line 23: is this possible to encapsulate into a common customDocumentLoader file rather than redefining frequently?

js/__fixtures__/did-key-ed25519.json
```diff
@@ -0,0 +1,612 @@
...
+      "application/did+json": {
+        "id": "#z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL",
+        "type": "JsonWebKey2020",
+        "controller": "did:key:z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL",
+        "publicKeyJwk": {
+          "crv": "Ed25519",
+          "x": "VDXDwuGKVq91zxU6q7__jLDUq8_C5cuxECgd-1feFTE",
+          "kty": "OKP"
+        },
+        "privateKeyJwk": {
```
kdenhartog, line 14: Apart of me thinks it's probably a bad idea to include private key info in a security-vocab definition because people will likely copy the style and think it's acceptable. Off the top of my head I don't know how else to deterministically handle tests without mocking private material (seeds/private keys). I'll have to get back to this with a suggestion on another way to go with this.

js/__tests__/p384/credentials-v2.test.js
```diff
@@ -0,0 +1,133 @@
...
+      documentUrl: url,
+      document: didDoc,
+    };
+  }
+
+  console.error(url);
+  throw new Error('Unsupported context ' + url);
+};
+
+// this test is broken, JsonWebSignature2020
```
kdenhartog, line 100: Should we skip it until JsonWebSignature2020 gets fixed?

kdenhartog (changes requested):
Thanks a ton for putting all the effort in! I suspect we have a few differing opinions on things that would could easily fix during a pair programming session. Happy to sit down and sync with you on this if it would be faster or to take it over to clean it up if you give me permissions and we reach consensus via this PR review process. In total this looks really close given the mammoth swath of effort and work you already did. Thanks for taking this on!

contexts/security-v3-unstable.jsonld
OR13: IIRC, tests confirmed that the definitions were not usable with it set to true. I suspect this is caused by the inability of some libraries to override the default security context, but you certainly can't mix the contexts with this set.

if we set this to true, the file will be unusable unless you can override `sec/v2 -> sec/v3-unstable` in your document loader.

js/__fixtures__/contexts/credentials-v2-unstable.json
OR13: the current vc context defines suites: https://github.com/w3c/vc-data-model/blob/master/contexts/credentials/v1#L163

I think we need to do the following:

1. define suites in their own context, for example bbs-v1.json
2. define suites in the security context, for example security/v3-unstable.json
3. define suites in the vc context, for example credentials/v2

yes, keeping things in sync will suck without tests... I propose we write tests.

I don't know if the vc data model would accept an external context like sec/v3... IIRC that is the reason things look the way they do today, people wanted a single file, with no external contexts.

js/__fixtures__/contexts/did-v1.json
OR13: this comment belongs on "did spec registries" :) I am just copying what is in https://github.com/w3c/did-spec-registries/blob/master/contexts/did-v1.jsonld

OR13: I agree btw, but a similar comment to the vc data model applies, both W3C DID and W3C VC specs each have their own context files, and I suspect we won't get them to agree to resolving an unstable context in maintained by the CCG.

js/__fixtures__/contexts/example-v1.json
OR13: yes, this is a copy paste, I like to know exactly what contexts are being used, I would suggest we remove the "example context" when we ship v2 credentials, its terrible.

js/__fixtures__/contexts/odrl.json
OR13: this file a dependency of examples/v1 see https://github.com/w3c-ccg/security-vocab/pull/80#discussion_r541010434

js/__fixtures__/credentials/bbs/verifiableCredential.json
OR13: yes, example dids are better imo.

js/__tests__/bbs/credentials-v2.test.js
OR13: yes, I think its probably wise to do that now.

js/__fixtures__/did-key-ed25519.json
OR13: https://tools.ietf.org/html/rfc7517#appendix-A.2

I think its a mistake NOT to include private keys, when trying to explain how cryptography works.

OR13:
@kdenhartog thanks for the review, let me take another pass on this.

contexts/security-v3-unstable.jsonld
OR13: thank you for pointing this out.

js/__fixtures__/credentials/bbs/credentialTemplate.json
OR13: I am keeping this until the correct one is provisioned, I also find the date style versioning to be less than intuitive....

js/__fixtures__/credentials/bbs/deriveProofFrame.json
OR13: same as before, until there is a real URL, no reason to change this.

js/__fixtures__/credentials/bbs/verifiableCredential.json
OR13: switched to examples

js/__tests__/bbs/credentials-v2.test.js
OR13: done

js/__fixtures__/did-key-ed25519.json
OR13: I removed the did key test vectors from the project, but I refuse to remove private keys from tests, if they are needed to write the test :)

js/__tests__/p384/credentials-v2.test.js
OR13: this was an error / related to the protected term thing, which I have resolved.

js/__fixtures__/contexts/odrl.json
OR13: I removed these, so we only rely on citizenship

js/__fixtures__/contexts/example-v1.json
OR13: I removed this

js/__fixtures__/contexts/did-v1.json
OR13: I removed this

contexts/security-v3-unstable.jsonld
OR13: fixed

OR13:
@kdenhartog I took another pass, removing a bunch of unnecessary things, and tightening up the definitions, ci tests are still passing.

.github/workflows/main.yml
```diff
@@ -1,42 +1,25 @@
...
-          CI: true
+    runs-on: ubuntu-latest
 
-      - name: Test
-        run: yarn test --ci --coverage --maxWorkers=2
-        env:
-          CI: true
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
```
davidlehn, line 12:
```suggestion
        node-version: [10.x, 12.x, 14.x]
```

contexts/security-v3-unstable.jsonld
```diff
@@ -321,6 +379,7 @@
     "signature": "https://w3id.org/security#signature",
     "signatureAlgorithm": "https://w3id.org/security#signingAlgorithm",
     "signatureValue": "https://w3id.org/security#signatureValue",
+    "proofValue": "https://w3id.org/security#proofValue",
```
davidlehn, line 382: Move up into alpha order.

OR13: won't fix without lint automation, @tplooker does your linter support sorting fields?

OR13:
@kdenhartog can you review, make clear what additional changes are requested if any?

contexts/security-v3-unstable.jsonld
gjgd: You can use:
- https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines if you use VSCode
- `:sort` command if you use Vim

OR13: https://github.com/w3c-ccg/security-vocab/issues/83

js/__fixtures__/contexts/credentials-v2-unstable.json
kdenhartog: In my opinion, I'd prefer to see us drop these terms from the V2-unstable context and start heading in the direction where only definitions related to that vocab / spec are defined in that context. I'd live with option 1 as well, but I think that goes against what the JSON only folks were after, so we'd probably need to make sure it doesn't present any problems for them as well, but that's got to be discussed in the VC maintenance WG.

js/__fixtures__/controllers/ed25519.json
```diff
@@ -0,0 +1,13 @@
+{
+    "@context": ["https://www.w3.org/ns/did/v1"],
```
kdenhartog, line 2: Can you locally cache this context as well?

js/__fixtures__/keys/bls12381-g2.json
```diff
@@ -0,0 +1,7 @@
+{
+    "id": "did:example:abc#456",
+    "type": "Bls12381G2Key2020",
+    "controller": "did:example:abc",
+    "publicKeyBase58": "26d2BdqELsXg7ZHCWKL2D5Y2S7mYrpkdhJemSEEvokd4qy4TULJeeU44hYPGKo4x4DbBp5ARzkv1D6xuB3bmhpdpKAXuXtode67wzh9PCtW8kTqQhH19VSiFZkLNkhe9rtf3",
+    "privateKeyBase58": "7DPkeiuhdQGGLZrbaJ41yh8HaCEPTVKBMuS5t9g8xVGH"
```
kdenhartog, line 6: I like how you separated out the private keys from the did documents in the controller folder. This seems like a good compromise.

js/__tests__/bls12381-g2/credentials-v2.test.js
```diff
@@ -0,0 +1,68 @@
...
+      suite,
+      documentLoader
+    });
+
+    const credentialVerified = await vcld.verifyCredential({
+      credential: {...verifiableCredential},
+      suite: new BbsBlsSignature2020({}),
+      documentLoader,
+    });
+    // console.log(JSON.stringify(verifiableCredential, null, 2))
```
kdenhartog, line 43: Does this need to be kept?

js/__tests__/bls12381-g2/credentials-v2.test.js
```diff
@@ -0,0 +1,68 @@
...
+    const deriveProofFrame = require('../../__fixtures__/credentials/bls12381-g2/deriveProofFrame.json');
+    const derivedProofCredential = await deriveProof(
+      verifiableCredential,
+      deriveProofFrame,
+      {
+        suite: new BbsBlsSignatureProof2020(),
+        documentLoader,
+      }
+    );
+    // console.log(JSON.stringify(derivedProofCredential, null, 2))
```
kdenhartog, line 58: same here

js/__tests__/bls12381-g2/credentials-v2.test.js
```diff
@@ -0,0 +1,68 @@
...
+    const credentialVerified = await vcld.verifyCredential({
+      credential: {...derivedProofCredential},
+      suite: new BbsBlsSignatureProof2020({}),
+      documentLoader,
+    });
+    // console.log(JSON.stringify(credentialVerified, null, 2))
```
kdenhartog, line 64: and here

js/__tests__/ed25519/credentials-v2.test.js
```diff
@@ -0,0 +1,42 @@
...
+    documentLoader
+  });
+
+  const result = await vcld.verifyCredential({
+    credential: verifiableCredential,
+    suite: new Ed25519Signature2018(),
+    documentLoader,
+  });
+
+  // console.log(JSON.stringify(verifiableCredential, null, 2))
```
kdenhartog, line 40: can be dropped

js/__tests__/p384/credentials-v2.test.js
```diff
@@ -0,0 +1,46 @@
...
+      suite,
+      documentLoader
+    });
+
+    const result = await vcld.verifyCredential({
+      credential: verifiableCredential,
+      suite: new JsonWebSignature(),
+      documentLoader,
+    });
+    // console.log(JSON.stringify(verifiableCredential, null, 2))
```
kdenhartog, line 43: https://media.giphy.com/media/3o7WTxyMSVN7lM5I7C/giphy.gif

js/__tests__/p384/security-v2.test.js
```diff
@@ -0,0 +1,42 @@
...
+it('can sign and verify', async () => {
+  const key = await JsonWebKey.from(require('../../__fixtures__/keys/p384.json'));
+  const suite = new JsonWebSignature({
+    key,
+  });
+
+  const signedDocument = await jsigs.sign(
+    {
+      '@context': ['https://w3id.org/security/v3-unstable', {
+        schema: 'http://schema.org/',
```
kdenhartog, line 21: Ahh smart way to reduce loading the whole schema.org context. The method I've normally gone with is just caching one of the files downloadable here: https://schema.org/docs/developers.html

kdenhartog (commented):
There's a few minor nits I noticed, but in general I think this is good enough to be merged. I think the best way to handle the VC/DID repeat terms issues is by merging this for now, solving where appropriate and coming back to update this later rather than holding this work up.

contexts/security-v3-unstable.jsonld
```diff
@@ -1,13 +1,71 @@
...
     "type": "@type",
     "@protected": true,
     "JsonWebKey2020": {
       "@id": "https://w3id.org/security#JsonWebKey2020"
     },
     "JsonWebSignature2020": {
-      "@id": "https://w3id.org/security#JsonWebSignature2020"
+      "@id": "https://w3id.org/security#JsonWebSignature2020",
+      "@context": {
+        "@version": 1.1,
```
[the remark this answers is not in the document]
OR13: my intention in this PR was not to style the contexts, but shows they don't result in errors. I opened https://github.com/w3c-ccg/security-vocab/issues/84 to track removing the verbosity

OR13:
@kdenhartog 

>On one hand the JSON only folks don't want to have to learn the dirty details of contexts and on the other hand we need to be properly separating the focus of the contexts in order to handle the variety of changes to these contexts.

Usually, I would agree, but the VC Data Model is about a semantic model for credentials and `@context` is required..... so really, we should be helping everyone experience as little JSON-LD as possible, but not hinting that they are not responsible for understanding the functionality that accompanies required terms.

OR13:
I removed the comments, thanks @kdenhartog for these awesome CRs!

js/__fixtures__/controllers/ed25519.json
OR13: this a good catch, the document loader should be 100% explicit, and we should cache locally 100% of contexts.

OR13: I have cached it locally (it was cached locally already, but better to be explicit with the document loader.

kdenhartog (approved):
While we haven't found a great solution that we'll all agree to for context splitting yet, this PR addresses the concerns it's intended to and looks good in my opinion. Let's go ahead and merge this and we can make changes to the contexts as we work through the best way that everyone agrees to later.

OR13:
@kdenhartog yes, we should not attempt to tackle context splitting without ci tests in place :)

OR13:
@tplooker can you review and merge?

PR #82, opened 2020-12-02
fix: Fix inconsistent 'proofPurpose' term definitions.

peacekeeper:
Signed-off-by: Markus Sabadello <markus@danubetech.com>

### DIFF
--- a/contexts/security-v3-unstable.jsonld
+++ b/contexts/security-v3-unstable.jsonld
@@ -33,6 +33,7 @@
           "@id": "https://w3id.org/security#proofPurpose",
           "@type": "@vocab",
           "@context": {
+            "@version": 1.1,
             "@protected": true,
             "id": "@id",
             "type": "@type",
@@ -112,6 +113,7 @@
           "@id": "https://w3id.org/security#proofPurpose",
           "@type": "@vocab",
           "@context": {
+            "@version": 1.1,
             "@protected": true,
             "id": "@id",
             "type": "@type",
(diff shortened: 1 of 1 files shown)


OR13 (approved):

OR13:
@peacekeeper @dlongley I am triggered by the `version` repetition here... is this really necessary? and if so why not apply it to every context?

peacekeeper:
@OR13 In the [current file](https://github.com/w3c-ccg/security-vocab/blob/master/contexts/security-v3-unstable.jsonld) some definitions of `proofPurpose` have the `@version` statement, others don't. Therefore I'm getting the error again about redefining protected terms. This PR makes sure all `proofPurpose` definitions are the same everywhere by adding `@version` to those that don't have it.

Maybe an equivalent fix would be to remove `@version` everywhere, not sure, I haven't tried that. But the current file which has it in some definitions but not in others is clearly broken...

dlongley:
@OR13, @peacekeeper -- what I believe needs to be done is the removal of checking `@version` when comparing term definitions in JSON-LD 1.1 processors because the term definitions are "the same" whether or not it is present (which is what the spec calls for). Once that's done, we wouldn't need to see `@version` in any of these definitions, it isn't necessary. But also, whether or not it was there would not affect the protection rules checking.

OR13:
@dlongley I think you are saying that a code change should happen, before the version is removed, but until then, version should be everywhere.

dlongley:
Quick update here -- code changes to address this are in the works right now. We do not want `@version` proliferating, it's a barrier to adoption and creates confusion.

OR13:
@tplooker ^ cross linking https://github.com/mattrglobal/jsonld-lint/issues/20

PR #95, opened 2021-04-09
Add EcdsaSecp256k1RecoverySignature2020

clehner:
And `EcdsaSecp256k1RecoveryMethod2020`.

Specification:
https://identity.foundation/EcdsaSecp256k1RecoverySignature2020/
https://github.com/decentralized-identity/EcdsaSecp256k1RecoverySignature2020

Cryptographic suite registered:
https://w3c-ccg.github.io/ld-cryptosuite-registry/#ecdsasecp256k1recoverysignature2020
https://w3c-ccg.github.io/ld-cryptosuite-registry/#ecdsasecp256k1recoverymethod2020

DID verification method type registered:
https://www.w3.org/TR/did-spec-registries/#ecdsasecp256k1recoverymethod2020

For the JSON-LD context, I did not use the context file or IRIs from the specification,
but instead copied what is already here for `JsonWebKey2020`/`JsonWebSignature2020`, defining new IRIs under the security context.

### DIFF
--- a/index.html
+++ b/index.html
@@ -167,6 +167,29 @@ until it is ratified as an official document via the World Wide Web Consortium.
         </pre>
       </section>
 
+      <section id="EcdsaSecp256k1RecoverySignature2020" about="https://identity.foundation/EcdsaSecp256k1RecoverySignature2020#EcdsaSecp256k1RecoverySignature2020"
+      typeof="rdfs:Class">
+        <h3>EcdsaSecp256k1RecoverySignature2020</h3>
+        <p>
+        This class represents a linked data signature suite. See <a href="https://w3c-ccg.github.io/ld-cryptosuite-registry/#ecdsasecp256k1recoverysignature2020">ecdsasecp256k1recoverysignature2020</a>.
+        </p>
+        <dl>
+          <dt>Status</dt>
+          <dd property="vs:term_status">stable</dd>
+          <dt>Expected properties</dt>
+          <dd>type, created, verificationMethod, proofPurpose, jws</dd>
+        </dl>
+        <pre class="example prettyprint language-jsonld">
+{
+  "type": "EcdsaSecp256k1RecoverySignature2020",
+  "created": "2020-04-11T21:07:06Z",
+  "verificationMethod": "did:example:123#vm-3",
+  "proofPurpose": "assertionMethod",
+  "jws": "eyJhbGciOiJFUzI1NkstUiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..pp9eiLCMfN4EfSB3cbl3UxJ4TtgUaTfByDaaB6IZbXsnvIy5AUIFjbgaiFNtq9-3f8mP7foD_HXpjrdWZfzlwAE"
+}
+        </pre>
+      </section>
+
       <section id="EcdsaSecp256k1VerificationKey2019" about="https://w3id.org/security#EcdsaSecp256k1VerificationKey2019"
       typeof="rdfs:Class">
         <h3>EcdsaSecp256k1VerificationKey2019</h3>
@@ -197,6 +220,42 @@ until it is ratified as an official document via the World Wide Web Consortium.
 
 
 
+      <section id="EcdsaSecp256k1RecoveryMethod2020" about="https://identity.foundation/EcdsaSecp256k1RecoverySignature2020#EcdsaSecp256k1RecoveryMethod2020"
+      typeof="rdfs:Class">
+        <h3>EcdsaSecp256k1RecoveryMethod2020</h3>
+        <p>
+          This class represents a linked data signature verification key. See <a href="https://w3c-ccg.github.io/ld-cryptosuite-registry/#ecdsasecp256k1recoverymethod2020">ecdsasecp256k1recoverymethod2020</a>.
+          </p>
+          <dl>
+            <dt>Status</dt>
+            <dd property="vs:term_status">stable</dd>
+            <dt>Expected properties</dt>
+            <dd>id, type, controller</dd>
+          </dl>
+<pre class="example prettyprint language-jsonld" title="Example using publicKeyJwk">
+{
+  "id": "did:example:123#WqzaOweASs78whhl_YvCEvj1nd89IycryVlmZMefcjU",
+  "type": "EcdsaSecp256k1RecoveryMethod2020",
+  "controller": "did:example:123",
+  "publicKeyJwk": {
+    "crv": "secp256k1",
+    "x": "4xAbUxbGGFPv4qpHlPFAUJdzteUGR1lRK-CELCufU9w",
+    "y": "EYcgCTsff1qtZjI9_ckZTXDSKAIuM0BknrKgo0BZ_Is",
+    "kty": "EC",
+    "kid": "WqzaOweASs78whhl_YvCEvj1nd89IycryVlmZMefcjU"
+  }
+}
+          </pre>
+
+<pre class="example prettyprint language-jsonld" title="Example using blockchainAccountId">
+{
+  "id": "did:example:123#keys-1",
+  "type": "EcdsaSecp256k1RecoveryMethod2020",
+  "controller": "did:example:123",
+  "blockchainAccountId": "0x89a932207c485f85226d86f7cd486a89a24fcc12@eip155:1"
+}
+</pre>
+      </section>
 
 
       <section id="RsaSignature2018" about="https://w3id.org/security#RsaSignature2018"


contexts/security-v3-unstable.jsonld
```diff
@@ -67,6 +67,69 @@
         }
       }
     },
+    "EcdsaSecp256k1RecoveryMethod2020": {
+      "@id": "https://w3id.org/security#EcdsaSecp256k1RecoveryMethod2020"
+    },
+    "EcdsaSecp256k1RecoverySignature2020": {
+      "@id": "https://w3id.org/security#EcdsaSecp256k1RecoverySignature2020"
+      "@context": {
+        "@version": 1.1,
```
msporny, line 76:
```suggestion
```

JSON-LD 1.1 does feature detection, so you don't need to declare the version anymore.

contexts/security-v3-unstable.jsonld
```diff
@@ -67,6 +67,69 @@
...
+        "created": {
+          "@id": "http://purl.org/dc/terms/created",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "domain": "https://w3id.org/security#domain",
+        "expires": {
+          "@id": "https://w3id.org/security#expiration",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "jws": "https://w3id.org/security#jws",
```
msporny, line 90:
```suggestion
        "jws": "https://w3id.org/security#jws",
```

Suggest you switch over to using multibase for the signature... unless you really want to continue using JOSE? If you continue using JOSE, the security suite will have to effectively make every non-secp256k1 parameter illegal and enforce it. I suggest that is a bad idea and will lead to larger attack surface and misimplementations (people will just pull in JOSE libraries with defaults (that support everything) and go with it).

msporny (changes requested):
We should probably start moving away from kitchen-sink security contexts to more surgically focused security contexts... For example:

https://github.com/digitalbazaar/ed25519-signature-2020-context/blob/master/contexts/ed25519-signature-2020-v1.jsonld

Example of usage here:

https://github.com/w3c/did-test-suite/pull/54/files#diff-c81ddb975aa4caf0733a08c11a8b2f8bde67286e56b24223abd77cc27fa587d4R13-R17

We may want to start enforcing this now so we don't keep going down a path that is a dead end. /cc @peacekeeper @OR13 @dlongley

OR13:
I suggest only adding https://w3id.org/#EcdsaSecp256k1RecoverySignature2020 to the html so its documented _here_ and then setup a on w3id.org to an context file for just that suite as manu mentioned.

examples: https://github.com/perma-id/w3id.org/pulls?q=is%3Apr+author%3AOR13+is%3Aclosed

OR13:
@msporny @dlongley how will we handle jsonld-signatures without sec-v3 or similar, do you think folks will just document all the contexts their libraries support?

I agree with we should make it possible for folks to pull only what they need, but we should also make it easy to implement VCs and LD Proofs and I'm not sure how you plan to move beyond sec-v2 in your libraries....

I would suggest that we keep using kitchen sinks for that... but that they load each of the suites your tooling supports... I know I argued against that with @kdenhartog but I think its the only way to make those tools reasonable.

msporny:
> I would suggest that we keep using kitchen sinks for that...

Yes, you raise a good point... which is that we want an easy on-ramp for people using this technology... and kitchen sinks certainly help there. My concern is that those kitchen sinks keep being used in production (and they will).

I'm torn on this one... will think about it more.

OR13:
@msporny I think this is not actually a big deal.

sec-v3 ~= module/version/3/index.js

all it should do is export the suites and terms that v3 supports... from their respective contexts.

At some point in the future everyone will upgrade from sec-v2 to sec-v3... and hopefully nothing breaks....

The more often we as a community do this, they more confident we are in JSON-LD Signatures.

Developers need to know which version of credentials they support and which version of ld signatures they support.

We should make this very clear.

dlongley:
@dmitrizagidulin,

Do you mind chiming in here on @OR13's question (quoted below) -- based on what we've been doing with the latest updates to jsigs and related suites and what the plan is there?

>  how will we handle jsonld-signatures without sec-v3 or similar, do you think folks will just document all the contexts their libraries support?

contexts/security-v3-unstable.jsonld
clehner: To be sure that this is okay... are we using any features that could be misinterpreted by JSON-LD 1.0 processors?

<https://w3c.github.io/json-ld-api/#features> says:

> Publishers are encouraged to use the `@version` map entry within a context set to 1.1 to ensure that JSON-LD 1.0 processors will not misinterpret JSON-LD 1.1 features.

TallTed: @manu - I can find nothing in the [JSON-LD 1.1 Rec](https://www.w3.org/TR/json-ld11/) about feature detection, and `"@version": 1.1` in many examples there. Please link to the new optionality?

dlongley: @TallTed, it's more that all version 1.1 features are defined such that they are either compatible or would trigger an error if a processing using 1.0 mode hits them. The `@version` tag just makes that happen immediately instead of when the feature is hit. The JSON-LD WG decided that explicitly declaring `@version` was a bad idea and harms adoption, and made an effort to ensure that you can just use the features you want/need and, if they are incompatible with JSON-LD 1.0, then such a processor will throw an error.

contexts/security-v3-unstable.jsonld
clehner: I've opened an issue on the specification about using proofValue and multibase:
https://github.com/decentralized-identity/EcdsaSecp256k1RecoverySignature2020/issues/12

>  If you continue using JOSE, the security suite will have to effectively make every non-secp256k1 parameter illegal and enforce it.

I read this as referring to `publicKeyJwk`, which is already defined as `@json` but the suite specification could add more guidance and/or requirements about it, so I've opened this issue:
https://github.com/decentralized-identity/EcdsaSecp256k1RecoverySignature2020/issues/13
Or did you mean about the JOSE header in the signature? I'm not sure what would be non-secp256k1 parameters there.

clehner:
@OR13 @msporny @dlongley Thanks for the pointers.

I've removed the changes to the security context file. I also reverted to using the original type IRIs defined in the suite specification, for consistency with existing usage and as it is in [did-spec-registries](https://w3c.github.io/did-spec-registries/#ecdsasecp256k1recoverymethod2020).

I also opened a PR to add a redirect for `w3id.org/security/suites/secp256k1recovery-2020/v1` to the existing JSON-LD context file, although I am not sure if this is still needed: https://github.com/perma-id/w3id.org/pull/2108

contexts/security-v3-unstable.jsonld
TallTed: @dlongley - As I recall from my (unfortunately brief) involvement with the JSON-LD WG working on the 1.1 spec, the `@version` declaration was necessary to make relative URIs behave properly for some use cases (including at least one that concerned me and OpenLink), and the lack of that declaration did not trigger any processing error in 1.0-based processors but did break the functionality which was making use of relative URIs.

I am now quite concerned that the hard-won solution to that issue has apparently been discarded, and more that people who clearly have no problem including the `@version` declaration are being advised not to do so -- when it remains at least legal, and again, clearly is not blocking their adoption/use of JSON-LD 1.1.

OR13: I guess I need to write some tests for this case... does VC Data Model require JSON-LD 1.1? does DID Core?

msporny:
@clehner I'm seeing merge conflicts on this PR, can you please rebase?

@clehner, are we good here -- have everyone's concerns been addressed?

index.html
```diff
@@ -196,6 +219,42 @@ <h3>EcdsaSecp256k1VerificationKey2019</h3>
...
+  }
+}
+          </pre>
+
+<pre class="example prettyprint language-jsonld" title="Example using blockchainAccountId">
+{
+  "id": "did:example:123#keys-1",
+  "type": "EcdsaSecp256k1RecoveryMethod2020",
+  "controller": "did:example:123",
+  "blockchainAccountId": "0x89a932207c485f85226d86f7cd486a89a24fcc12@eip155:1"
```
msporny, line 254: Ideally, this would be a URI... I thought that was the whole point of `blockchainAccountId`?

msporny (approved):
Overall, LGTM for registration.

contexts/security-v3-unstable.jsonld
dlongley: @TallTed -- that sounds like its getting off topic. But there's errata for JSON-LD 1.0 around prefix / relative URI definitions -- and I believe there are generally other mechanisms that (may have been invented after your time with the WG) can be used to force the behavior you want without having to resort to using `@version`. For example, using the new `@prefix` keyword, which, IIRC, will also trigger a JSON-LD 1.0 error. Again, I'm not sure which case you're talking about specifically, but my memory is that the group made it possible to trigger errors with JSON-LD 1.0 by using JSON-LD 1.1 specific features (for whatever specific problem you wanted solved) instead of having to rely on a catch-all version declaration.

TallTed: @dlongley -- Our issue involved relative URIs as property identifiers being expanded by some concatenation rule, rather than following [RFC 3986 rules](https://datatracker.ietf.org/doc/html/rfc3986), when expanding relative to `@base`. See [(long!) discussion](https://github.com/json-ld/json-ld.org/issues/488).  Assuming the Playground is fully up-to-date for JSON-LD 1.1, it appears to me that later revisions to that spec re-broke the things we had gotten to work earlier.  I'm checking with others to confirm my read on this, and then we'll log issues as appropriate.

PR #97, opened 2021-04-16
feat: add x509 properties

tplooker:
Adds the ability to express the relationship between a verification method used to produce a Linked Data Proof and an X.509 based certificate.


index.html
```diff
@@ -2484,6 +2484,84 @@ <h3>serviceEndpoint</h3>
                   </pre>
       </section>
 
+      <section id="x509Certificate" about="https://w3id.org/security#x509Certificate"
+      typeof="rdf:Property">
+        <h3>x509Certificate</h3>
+        <p>
+          The x509Certificate property is used to associate an <a href="https://tools.ietf.org/html/rfc5280">X.509 Certificate</a> with a proof.
```
msporny, line 2491: This property should specify the encoding format for the value. I suggest we use multibase.

index.html
```diff
@@ -2484,6 +2484,84 @@ <h3>serviceEndpoint</h3>
...
+                      }
+                    }
+                  </pre>
+      </section>
+
+      <section id="x509CertificateFingerprint" about="https://w3id.org/security#x509CertificateFinger"
+      typeof="rdf:Property">
+        <h3>x509CertificateFingerprint</h3>
+        <p>
+          The x509CertificateFingerprint property is used to associate an <a href="https://tools.ietf.org/html/rfc5280">X.509 Certificate</a> with a proof via its SHA-256 based fingerprint.
```
msporny, line 2538: This property should specify the encoding format for the value. I suggest we use multibase.

msporny (changes requested):
We should specify the encoding format for the values, ideally using multibase.

index.html
David-Chadwick: Since X.509/PKIX already specifies fingerprint formats in Ascii wouldnt it be better to simply use these?

dlongley: @David-Chadwick, if they use multibase, they are compressible in CBOR-LD.

contexts/security-v3-unstable.jsonld
```diff
@@ -84,87 +84,93 @@
...
+            "@type": "@id"
+          }
         }
+      },
+      "x509Certificate": {
+        "@id": "https://w3id.org/security#X509Certificate"
+      },
+      "x509CertificateFingerprint": {
+        "@id": "https://w3id.org/security#X509CertificateFingerprint"
       }
```
dlongley, line 173:
```suggestion
      "x509CertificateFingerprint": {
        "@id": "https://w3id.org/security#X509CertificateFingerprint",
        "@type": "https://w3id.org/security#multibase"
      }
```

contexts/security-v3-unstable.jsonld
```diff
@@ -84,87 +84,93 @@
...
+          "proofValue": "https://w3id.org/security#proofValue",
+          "verificationMethod": {
+            "@id": "https://w3id.org/security#verificationMethod",
+            "@type": "@id"
+          }
         }
+      },
+      "x509Certificate": {
+        "@id": "https://w3id.org/security#X509Certificate"
+      },
```
dlongley, line 170: Can simplify to this, right?

```suggestion
      "x509Certificate": "https://w3id.org/security#X509Certificate",
```

index.html
David-Chadwick: You should think of the customers/users. If the person already has an X.509 PKC along with a fingerprint then you should accept it as is, without requiring him to jump through formatting hoops to convert it into something that you want.

dlongley: Priority of constituencies: 1. end users, 2. devs, 3. spec/vocab authors.

I think the "users" you're referring to here are "devs". If a use case is consumption of this information by end users via QR codes (i.e., made possible by CBOR-LD), then we should favor multibase encoding because end users are the top priority. If that's not at all expected to be a common use case, then we should favor devs next -- and if they would significantly prefer another format we should go with that -- so I'd be in agreement with you then. So, what do the expected common use cases suggest we should do?

msporny: @David-Chadwick, we can't easily compress the ASCII fingerprints in PKIX, and the driving use case for this feature is the use in vaccination certificates, which need to be highly byte compressed using multibase and CBOR-LD. You can always generate a human-readable fingerprint from the compressed data.

To give you an idea of the size difference, an ASCII SHA-256 x509 certificate fingerprint is 95 bytes, whereas a byte encoding of the same value is 32 bytes (which is 1/3rd of the size). This matters when you want to keep payload sizes below 500 bytes (which is the typical top end scanning range of most low-end QR Code scanners).

David-Chadwick: I think compressing any VC into a QR code is a non-starter for several reasons. 1. It breaks privacy and does not allow for selective disclosure 2. It constrains the VCs contents, esp. photos are ruled out. The WHO prefers the QR code to be a pointer to a digital twin as do many other implementors.

David-Chadwick: @dlongley. Yes I am thinking of devs (or implementors) who are used to using X.509 tools.

tplooker: > The WHO prefers the QR code to be a pointer to a digital twin as do many other implementors.

As a member of the WHO SVC working group I can assure you that this claim is completely incorrect.

contexts/security-v3-unstable.jsonld
tplooker: Yeap was just following the convention further up in the file but agree we can shorthand it

index.html
tplooker: Agreed, however because multi-base provides a surface for many choices of encoding we should consider providing implementation guidance around usage of the property in the linked data proof specification (e.g please consider using x encoding via multi-base)

msporny: > we should consider providing implementation guidance around usage of the property in the linked data proof specification

Yes, agreed... we may even want to do that in the vocabulary? I expect us to suggest base64pad (which I believe is the default for X509 certs?)... other options to limit to might be base64url or base58?

tplooker: For the certificate I was thinking we could define it as base64url then the PEM conversion is trivial?

tplooker: For instance the certificate given in the example is a just the base64url contents from a PEM encoded cert

index.html
David-Chadwick: then please explain how you can reconcile your assertion with this statement from the WHO:
"Subsequent vaccine administration events logged to the same SVC paper card will be added as new entries to the signed SVC document on the SVC Registry that is referenced by the SVC paper card’s UVCI."
A static QR code containing a subset of initial event information cannot fulfil this requirement. Only a QR code that points to the digital twin can (unless you know of some other mechanism)

tplooker: Due to the confidentiality agreement signed as apart of that working group members cannot discuss details that are not apart of public releases until they are.

> Subsequent vaccine administration events logged to the same SVC paper card will be added as new entries to the signed SVC document on the SVC Registry that is referenced by the SVC paper card’s UVCI

This statement simply does not equate to your assertion of "The WHO prefers the QR code to be a pointer to a digital twin" but as clarity please refer to section 3.3.2 of RC 1.

> Release Candidate 2 of this specification will provide more detail about possible options to print barcoded content on the SVC card besides just the UVCI.

Furthermore throughout the document there are numerous references to being able to perform an offline verification check of an SVC, something that cannot be done if the information encoded in the 2D barcode is only a reference. See section 3.4.2 for instance.

tplooker: Since x509 certificates in PEM format already leverage base64url are we best to use this encoding scheme for this property too?

index.html
msporny: > For the certificate I was thinking we could define it as base64url then the PEM conversion is trivial?

Well, PEM uses base64pad, not base64url...

I believe the current authoritative spec for certificate encoding is [RFC7468](https://tools.ietf.org/html/rfc7468)... which doesn't use base64url... it uses base64pad... with headers and trailers. See https://tools.ietf.org/html/rfc7468#section-2 for more information. 

I'm still torn about breaking rank with PEM encoding... which does specify `----- BEGIN` and `----- END` headers and trailers. If we do that, we can't compress down to binary for CBOR-LD use cases... but then again, we probably won't be compressing `x509Certificate` values in CBOR-LD. The same is not true for `x509CertificateFingerprint`.

So, we could either just follow the rules in https://tools.ietf.org/html/rfc7515#section-4.1.6 -- you'll want to make this a `@list` in the JSON-LD Context as order matters for `x509Certificate`. The upside here is it just does what JOSE does... the downside is that we won't be able to CBOR-LD compress the value (not that doing so is a regular use case?).

The other option is multibase-encode the PEM certificate, still use `@list` to be able to encode certificate chains, and don't support PEM header/trailers.

@dlongley @tplooker, thoughts?

index.html
msporny: > Since x509 certificates in PEM format already leverage base64url are we best to use this encoding scheme for this property too?

They use base64pad, not base64url. I expect we would default to base64pad.

index.html
dlongley: > The other option is multibase-encode the PEM certificate, still use @list to be able to encode certificate chains, and don't support PEM header/trailers.

There is already precedence via https://tools.ietf.org/html/rfc7515#section-4.1.6 for doing base64(pad)-encoded DER (i.e., stripping the PEM armor). So, at minimum we should be able to do an `@list` with values that have that encoding. However, I think we should take it a step further to enable compression in CBOR-LD by making each value multibase encoded (one extra character prefixed before each base64pad value).

tplooker:
@msporny @dlongley I have refactored this x509Certificate property is now x509CertificateChain and defined as `@list`, I have redefined it to use multi-base with the base64pad flavour. x509CertificateFingerprint -> x509CertificateThumbprint (cause naming things is hard :)), also changed this to be the SHA-1 thumbprint rather than SHA-256 to follow JOSE's conventions, we may need another property if we want to express SHA-256 too.

msporny:
> the x509Certificate property is now x509CertificateChain and defined as `@list`

+1

> I have redefined it to use multi-base with the base64pad flavour. 

+1

> x509CertificateFingerprint -> x509CertificateThumbprint (cause naming things is hard :))

-0.5 -- Why Microsoft decided to break ranks with the rest of the security industry and insist that we use "thumbprint" instead of "fingerprint" in JOSE... I don't know. What I do know is that:

The x509 IETF RFCs use the term "fingerprint" -- https://tools.ietf.org/html/rfc4210 (except for JOSE, of course)
OpenSSL uses the term "fingerprint" -- https://www.openssl.org/docs/man1.1.0/man1/openssl-x509.html
GPG uses the term "fingerprint" -- https://www.gnupg.org/gph/en/manual/r822.html
All Chromium-based browsers uses the term "fingerprint" -- https://ldapwiki.com/wiki/Certificate%20Fingerprint
The Wikipedia page uses the term "fingerprint" -- https://en.wikipedia.org/wiki/Public_key_fingerprint

Only Microsoft software seems to prefer "thumbprint" -- which is pretty `/` vs. `\`, `-` vs. `/` typical. :P

Can we please not use thumbprint? 

> also changed this to be the SHA-1 thumbprint rather than SHA-256 to follow JOSE's conventions, we may need another property if we want to express SHA-256 too.

We may want to consider using a Multibase-encoded Multihash value. If we don't do that, we'll want to use `x509Sha1CertificateFingerprint` and `x509Sha256CertificateFingerprint`.

I'd prefer we do the former... and then insist that it's base64pad + SHA-1 (or SHA-256) for now.

tplooker:
> Can we please not use thumbprint?

:) sure, lets use fingerprint I was on the fence anyway, COSE and JOSE have gone thumbprint and just about everyone else has gone fingerprint *sigh* this is why we can't have nice things.

> We may want to consider using a Multibase-encoded Multihash value. If we don't do that, we'll want to use x509Sha1CertificateFingerprint and x509Sha256CertificateFingerprint.

Im wary about encoding the semantics of something like the fingerprints hash into the value itself and relying on multi-hash to provide this distinction, any other approach though feels in-consistent so I've opted to go with it for now with a recommendation for SHA-1.

kdenhartog:
> @msporny @dlongley I have refactored this, the x509Certificate property is now x509CertificateChain and defined as `@list`, I have redefined it to use multi-base with the base64pad flavour. x509CertificateFingerprint -> x509CertificateThumbprint (cause naming things is hard :)), also changed this to be the SHA-1 thumbprint rather than SHA-256 to follow JOSE's conventions, we may need another property if we want to express SHA-256 too.

Agree with everything except the SHA-1 usage. IANA JOSE registry has made good efforts to prohibit usage of SHA1 in the Algorithms section and only leaves the SHA-1 thumbprint parameter for the sake of clarity that it existed at one point and because it's not actually an endorsement for a particular algorithm, more just a registration of a property name.

Instead, I'd suggest that we default x509CertificateThumbprint to SHA-256 instead of potentially fragmenting support over x509CertificateThumbprint and x509CertificateThumbprintSha256 since usage of SHA-1 should be deprecated at this point. This way we're agile enough to support current and potentially new algorithms in the future, but we're not being overly agile to the point where we're supporting deprecated algorithms. I'd be alright with saying we only support x509CertificateThumbprintSha256 as well so that we're setting good precedent for the next time we want to support a different algorithm when it comes to naming, but aren't having to defend the choice to support a deprecated algorithm on things that are new.

tplooker:
Unless we have a good reason for interop with X.509 I'm ok with changing the recommended hash function for the fingerprint to being SHA-256 @dlongley @msporny thoughts?

msporny:
> Unless we have a good reason for interop with X.509 I'm ok with changing the recommended hash function for the fingerprint to being SHA-256 @dlongley @msporny thoughts?

I'm fine w/ defaulting to SHA-256 -- but how do we signal that? One way is to assume a particular byte length means "SHA-2 256-bit output", the other is to explicitly name the property. I suggest we explicitly name the property -- `x509Sha256CertificateFingerprint`

Note -- we are going to regret this when SHA-256 is broken. :) -- at which point, we'll have wished we picked Multihash... I expect that to happen in a decade, and when that happens, we can go to `x509CertificateFingerprint` and use multihash + multibase.

kdenhartog:
> > Unless we have a good reason for interop with X.509 I'm ok with changing the recommended hash function for the fingerprint to being SHA-256 @dlongley @msporny thoughts?
> 
> I'm fine w/ defaulting to SHA-256 -- but how do we signal that? One way is to assume a particular byte length means "SHA-2 256-bit output", the other is to explicitly name the property. I suggest we explicitly name the property -- `x509Sha256CertificateFingerprint`
> 
> Note -- we are going to regret this when SHA-256 is broken. :) -- at which point, we'll have wished we picked Multihash... I expect that to happen in a decade, and when that happens, we can go to `x509CertificateFingerprint` and use multihash + multibase.
> 
> So, I guess my vote would be for `x509Sha256CertificateFingerprint` (which is always encoded in base64pad).

Multibase is definitely the better solution for long term support, but my hope (overly biased I'm sure) is that X.509 certificates are replaced by the time SHA-256 is broken 😄

tplooker:
> So, I guess my vote would be for x509Sha256CertificateFingerprint (which is always encoded in base64pad).

Wait now im confused I opted for x509CertificateFingerprint with a multihash then multibase encoded value as I thought that was what was prefered, happy to go with x509Sha256CertificateFingerprint as a multibase encoded value using base64pad instead though? Leaves us open to migrate to x509CertificateFingerprint in future?

David-Chadwick:
> Unless we have a good reason for interop with X.509'

Why are you even considering this is you have no intention of interop with people who use X.509 PKCs?

tplooker:
> Why are you even considering this is you have no intention of interop with people who use X.509 PKCs?

Sorry do you mean "if you have no"? I think you have misconstrued my comment, of course the whole intent of this PR is to establish interop with X.509 what I meant was there are a variety of choices in designing the integration that factor in other requirements that linked data proofs have such as compressibility via usage of schemes like multibase or multihash which effect the ease in which the interop with existing X.509 tools will occur

OR13 (approved):
I hope its not too late to add an example to did core.... who has that ball?

msporny:
> Wait now im confused I opted for x509CertificateFingerprint with a multihash then multibase encoded value as I thought that was what was prefered

Haha, I thought you were arguing against that!

Yes, this is what I would prefer: `x509CertificateFingerprint` that is a multihash value, which is multibase-encoded. 

Separately (in the vocabulary), we should insist that base64pad is the MUST supported multibase format with SHA-256 as the MUST supported hash format. That means that the base64pad decoded value should have a header of `0x4d1220` ... which means, matching against and then stripping the first 3 bytes should give you the remaining SHA-256 fingerprint bytes. 

Given that SHA-1 *signed* certificates have been banned since 2016 with a move to SHA-256, the industry is entirely capable of generating a SHA-256 fingerprint for certificates.

@kdenhartog note that SHA-1 certificate fingerprints aren't considered to be as broken as digital signatures using SHA-1 hashes... all that said, we really should get away from SHA-1 and enough time has passed since 2016 to insist that SHA-256 is used, AFAICT.

msporny:
> I hope its not too late to add an example to did core.... who has that ball?

It's not too late... and no one right now, PRs welcome. :)

dlongley:
> Yes, this is what I would prefer: x509CertificateFingerprint that is a multihash value, which is multibase-encoded.
> 
> Separately (in the vocabulary), we should insist that base64pad is the MUST supported multibase format with SHA-256 as the MUST supported hash format. That means that the base64pad decoded value should have a header of 0x4d1220 ... which means, matching against and then stripping the first 3 bytes should give you the remaining SHA-256 fingerprint bytes.

I agree with the above.

tplooker:
Ok I think we are getting closer, can I get another round of reviews? As I think the current text is what has been discussed in this PR.

tplooker:
Ping @msporny @dlongley @kdenhartog

index.html
```diff
@@ -2558,6 +2558,90 @@ <h3>serviceEndpoint</h3>
...
+                  </pre>
+      </section>
+
+      <section id="x509CertificateFingerprint" about="https://w3id.org/security#x509CertificateFingerprint"
+      typeof="rdf:Property">
+        <h3>x509CertificateFingerprint</h3>
+        <p>
+          The x509CertificateFingerprint property is used to associate an <a href="https://tools.ietf.org/html/rfc5280">X.509 Certificate</a> with a proof via its fingerprint.
+          The value is a <a href="https://tools.ietf.org/id/draft-multiformats-multihash-00.html#rfc.appendix.B.1">multihash</a> encoded then <a href="https://tools.ietf.org/id/draft-multiformats-multibase-00.html">multibase</a> 
+          encoded value using the <a href="https://tools.ietf.org/id/draft-multiformats-multibase-00.html#rfc.appendix.D.1">base64pad</a> varient. It is RECOMMENDED that the fingerprint value be the SHA-256
```
dlongley, line 2617:
```suggestion
          encoded value using the <a href="https://tools.ietf.org/id/draft-multiformats-multibase-00.html#rfc.appendix.D.1">base64pad</a> variant. It is RECOMMENDED that the fingerprint value be the SHA-256
```

index.html
```diff
@@ -2558,6 +2558,90 @@ <h3>serviceEndpoint</h3>
...
+          The following example demonstrates how a proof is expressed using
+          a JSON-LD proof with a link to an associated <code>x509CertificateFingerprint</code>:
+                  </p>
+                  <pre class="example prettyprint language-jsonld">
+                    {
+                      ...
+                      "proof": {
+                        "type": "Ed25519Signature2018",
+                        "created": "2020-04-03T18:03:33Z",
+                        "x509CertificateFingerprint": "egW8ixMpQ2W7SG46Y+ZzPkCUMjxmq7WIuMego9CZdhU=",
```
dlongley, line 2636: Note that I think this needs to start with `M` for base64pad.

index.html
```diff
@@ -2558,6 +2558,90 @@ <h3>serviceEndpoint</h3>
                   </pre>
       </section>
 
+      <section id="x509CertificateChain" about="https://w3id.org/security#x509CertificateChain"
+      typeof="rdf:Property">
+        <h3>x509CertificateChain</h3>
+        <p>
+          The x509CertificateChain property is used to associate a chain of <a href="https://tools.ietf.org/html/rfc5280">X.509 Certificates</a> with a proof. The value of this
+          property is an ordered list where each value in the list is an <a href="https://tools.ietf.org/html/rfc5280">X.509 Certificate</a> expressed as a DER PKIX format, that is encoded with <a href="https://tools.ietf.org/id/draft-multiformats-multibase-00.html">multibase</a> 
+          using the <a href="https://tools.ietf.org/id/draft-multiformats-multibase-00.html#rfc.appendix.D.1">base64pad</a> varient. The certificate directly associated to the verification method used to verify the proof MUST be the first element in the list. Subsequent
```
dlongley, line 2567:
```suggestion
          using the <a href="https://tools.ietf.org/id/draft-multiformats-multibase-00.html#rfc.appendix.D.1">base64pad</a> variant. The certificate directly associated to the verification method used to verify the proof MUST be the first element in the list. Subsequent
```

dlongley (approved):
LGTM, left a few nits.

index.html
OR13: https://tools.ietf.org/html/draft-multiformats-multibase-03

I might recommend base64url instead... which you be `u`.

contexts/security-v3-unstable.jsonld
```diff
@@ -706,6 +706,14 @@
       "@type": "@id"
     },
     "verifyData": "https://w3id.org/security#verifyData",
-    "wrappedKey": "https://w3id.org/security#wrappedKey"
+    "wrappedKey": "https://w3id.org/security#wrappedKey",
+    "x509CertificateChain": {
+      "@id": "https://w3id.org/security#X509Certificate",
```
msporny, line 711:
```suggestion
      "@id": "https://w3id.org/security#x509CertificateChain",
```

contexts/security-v3-unstable.jsonld
```diff
@@ -706,6 +706,14 @@
...
+      "@container": "@list"
```
msporny, line 712:
```suggestion
      "@type": "https://w3id.org/security#multibase",
      "@container": "@list"
```

index.html
msporny: > I might recommend base64url instead... which you be u.

We had suggested base64pad because that's what the default is for PEM encoding and for the `x509CertificateChain` field. I don't have a strong opinion either way, I expect folks that do base64pad also have base64url libraries available to them... and arbitrary base-encoding is fairly easy to implement.

The only rationale that I can think of for base64url by default is that we think it's going to end up in a URL somewhere. I don't expect it to, and if it does, I expect folks will URI-encode base64pad values?

The rationale for base64pad is because "that's what PEM-encoded x509 values are encoded as".

Leaning ever so slightly towards base64pad by default.... but again, no real strong feelings.

msporny (approved):
LGTM other than minor nits.

index.html
OR13: If the default string encoding is usable by just dropping the `M` that seems like a good enough justification, agree regarding URLs

tplooker: Updated now thanks

tplooker: I think continuing with base64pad for now to align better with PEM is the best approach, JOSE uses a mix of base64url and pad for x509 properties, so there is precedent there.

tplooker:
Multiple reviews, 3 approvals, merging.

PR #103, opened 2021-05-08
Add publicKeyMultibase definition.

msporny:
This PR partially addresses an issue in https://github.com/w3c/did-core/issues/707 by registering `publicKeyMultibase`, which is needed before registering `publicKeyMultibase` in https://github.com/w3c/did-spec-registries.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2100,7 +2100,10 @@ A <code>publicKeyHex</code> property is used to specify the hex-encoded version 
         <h3>publicKeyMultibase</h3>
         <p>
 The public key multibase property is used to specify the multibase-encoded
-version of a public key.
+version of a public key. Each key type definition is expected to specify
+one or more required encoding bases. Specifying only a single encoding
+base per key type is preferable as it reduces the burden to reach broad
+interoperability.
                   </p>
                   <dl>
                     <dt>Status</dt>
--- a/index.html
+++ b/index.html
@@ -69,7 +69,50 @@ companyURL: "https://mattr.global/"
           github: "https://github.com/w3c-ccg/security-vocab/",
 
           // extend the bibliography entries
-          //localBiblio: webpayments.localBiblio,
+          localBiblio: {
+            "ED25519-2020": {
+              title: "Ed25519 Signature 2020",
+              date: "November 2020",
+              href: "https://w3c-ccg.github.io/lds-ed25519-2020/",
+              authors: [
+                "Orie Steele"
+              ],
+              status: "Community Group Draft",
+              publisher: "W3C Credentials Community Group"
+            },
+            "LD-CRYPTOSUITE-REGISTRY": {
+              title: "The Linked Data Cryptosuite Registry",
+              date: "May 2021",
+              href: "https://w3c-ccg.github.io/ld-cryptosuite-registry/",
+              authors: [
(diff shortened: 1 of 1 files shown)


dlongley (approved):
Do we want to have a line in here indicating that key definitions may restrict which bases are supported?

msporny:
> Do we want to have a line in here indicating that key definitions may restrict which bases are supported?

Yeah, we do. I'm deep in something else right now, so change request would be good, if not, I may get back around to it.

index.html
```diff
@@ -2085,6 +2095,36 @@ <h3>publicKeyHex</h3>
         </pre>
       </section>
 
+      <section id="publicKeyMultibase" about="https://w3id.org/security#publicKeyMultibase"
+      typeof="rdf:Property">
+        <h3>publicKeyMultibase</h3>
+        <p>
+The public key multibase property is used to specify the multibase-encoded
+version of a public key.
```
dlongley, line 2103: Suggestion choice without the trailing advice:

```suggestion
version of a public key. Each key type definition is expected to specify
one or more required encoding bases.
```

index.html
```diff
@@ -2085,6 +2095,39 @@ <h3>publicKeyHex</h3>
...
 
+      <section id="publicKeyMultibase" about="https://w3id.org/security#publicKeyMultibase"
+      typeof="rdf:Property">
+        <h3>publicKeyMultibase</h3>
+        <p>
+The public key multibase property is used to specify the multibase-encoded
+version of a public key. Each key type definition is expected to specify
+one or more required encoding bases. Specifying only a single encoding
+base per key type is preferable as it reduces the burden to reach broad
+interoperability.
```
kdenhartog, line 2106: I'm a bit worried that there's quite a bit of tribal knowledge in this. For example, how do I encode a nist p256 key here? How do I specify the encoding. If this is meant to be documentation to be "hey this thing exists" then this is fine, but if we're actually expecting people to build interoperable software against this definition they'll undoubtedly get stuck.

For specific action items I'd like to see, I'd like to see us reference how encoding works by directly referencing [multibase](https://github.com/multiformats/multibase#multibase-table) and additionally reference how [multicodec table](https://github.com/multiformats/multicodec/blob/master/table.csv#L86) is used and finally include how serialization of the bytes should work as well. 

I'm all for this additional change, but I think this is a bit too light to actually be something that someone without prior knowledge could implement.

kdenhartog (changes requested):

index.html
OR13: yes, probably also a good idea to always recommend compressed form keys.

OR13: see also https://github.com/multiformats/multicodec/pull/210

msporny: I have applied suggestions from @kdenhartog, @peacekeeper, and @OR13 in https://github.com/w3c-ccg/security-vocab/pull/103/commits/924eea44439a31ea2f22b3180664931559df8c62.

If you want further changes, please suggest concrete spec text -- it's hard to read minds and then write spec text. :P

kdenhartog (approved):
Thanks for the additional details. This is good enough now to help people define properties and as we come back to defining suites with this properties we can further align it.

PR #109, opened 2021-05-19
Register missing type "X25519KeyAgreementKey2019"

OR13:
Related to https://github.com/perma-id/w3id.org/pull/2161

### DIFF
--- a/index.html
+++ b/index.html
@@ -593,6 +593,28 @@ and verified.
           </pre>
       </section>
 
+      <section id="X25519KeyAgreementKey2019" about="https://w3id.org/security#X25519KeyAgreementKey2019"
+      typeof="rdfs:Class">
+        <h3>X25519KeyAgreementKey2019</h3>
+        <p>
+          This class represents a verification key.
+          </p>
+          <dl>
+            <dt>Status</dt>
+            <dd property="vs:term_status">unstable</dd>
+            <dt>Expected properties</dt>
+            <dd>id, type, controller, publicKeyBase58</dd>
+          </dl>
+          <pre class="example prettyprint language-jsonld">
+{
+  "id": "did:example:123#z6LSeu9HkTHSfLLeUs2nnzUSNedgDUevfNQgQjQC23ZCit6F",
+  "type": "X25519KeyAgreementKey2019",
+  "controller": "did:example:123",
+  "publicKeyBase58": "96XGGce9y12gRTxEcnuvpP8tCZzE39WkpKHvUZ6DLg6d"
+}
+          </pre>
+      </section> 
+
       <section id="Ed25519VerificationKey2018" about="https://w3id.org/security#Ed25519VerificationKey2018"
       typeof="rdfs:Class">
         <h3>Ed25519VerificationKey2018</h3>


index.html
```diff
@@ -593,6 +593,28 @@ <h3>MerkleProof2019</h3>
           </pre>
       </section>
 
+      <section id="X25519KeyAgreementKey2019" about="https://w3id.org/security#X25519KeyAgreementKey2019"
+      typeof="rdfs:Class">
+        <h3>X25519KeyAgreementKey2019</h3>
+        <p>
+          This class represents a verification key.
```
msporny, line 600: It's more than just a verification key, right? We need to point out that it's a very specific type of verification key and explain that it's used for key agreement using Curve25519 -> X25519.

Also, needs to be dealt with in a separate PR, but isn't X25519KeyAgreementKey2020 (using publicKeyMultibase) missing?

msporny (changes requested):
The description of the key type needs to be updated.

index.html
OR13: @msporny I opened https://github.com/w3c-ccg/security-vocab/issues/111

OR13: > We need to point out that it's a very specific type of verification key and explain that it's used for key agreement using Curve25519 -> X25519.

No we don't, how you get an X25519 key is orthogonal to defining what it is.

For example, I might have gotten one by using a mnemonic or by directly generating it from randomness, in fact our did key implementation for X25519 by default does not generate it from Ed25519.... for example:

https://did.key.transmute.industries/did:key:z6LStaHNaMZzqpABWwzUumEZyqK7xT9dLBMi2tRp5FvVxLtW

In the DID Core WG we agreed to call keyAgreement a "verification relationship" built with "verification material":

https://w3c.github.io/did-core/#key-agreement
https://w3c.github.io/did-core/#verification-material

Can you propose language that you would prefer to see in stead of "This class represents a verification key."

I would assert that "This class represents a verification key."  is the least controversial thing we might say about a key that is used for key agreement, give it allows you to verify that ECDH is possible, and it makes clear you are talking about a public component.

OR13: I will say, I appreciate the pedantry around naming... its hard... but I also think this ship has essentially sailed thanks to the did core normative defs above... particularly "verification relationships" and "verification material"... I think we now have to align with that... or risk being viewed as undermining that work.

OR13:
Merging over the objection, so that the JSON-LD contexts will have a term definition:

https://w3id.org/security/suites/x25519-2019/v1

`https://w3id.org/security#X25519KeyAgreementKey2019`

Subsequent PRs welcome, but this term is widely used, and needs to have a definition.

msporny:
> Merging over the objection, so that the JSON-LD contexts will have a term definition:

I never objected to merging... I was merely saying: "We can do better than 'This class represents a verification key.'" <-- because that definition could apply to /any/ verification key. We should elaborate upon exactly what this particular class is used for... 'cause when a developer goes to read the definition, it won't help them understand exactly what it's for.

OR13:
@msporny I agree, let leave this open: https://github.com/w3c-ccg/security-vocab/issues/111

PR #115, opened 2021-07-09
Fix typo in section 3.33 domain

acoburn:
Section 3.33 is concerned with the `domain` property, but the text refers to the `challenge` property, which appears to be incorrect.


msporny (approved):
Great catch in some very old text. Thank you @acoburn !

Issue #125, opened 2021-12-17
replace all instances of "owner" with "controller"?


kkabrams:
should fix issue #114
I'm kind of guessing based on how some properties said they required id, context, and controller, but some examples had id, context, and owner.

ghost:
I guess this can be closed now, the pull request was merged.

PR #126, opened 2021-12-17
replace the usages of "owner" with "controller"

ghost:
fixes issue #125

### DIFF
--- a/index.html
+++ b/index.html
@@ -1178,11 +1178,11 @@ decryption, or digitally signing data.
             <dt>Status</dt>
             <dd property="vs:term_status">stable</dd>
             <dt>Expected properties</dt>
-            <dd>owner, privateKeyPem, publicKeyPem</dd>
+            <dd>controller, privateKeyPem, publicKeyPem</dd>
           </dl>
         <p>
 The example below describes a cryptographic key that contains both the
-public and private key as well as the owner of the key. The owner property
+public and private key as well as the controller of the key. The controller property
 is described in the <a href="https://w3id.org/commerce">Commerce Vocabulary</a>.
         </p>
 <pre class="example prettyprint language-jsonld">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1702,14 +1702,14 @@ <h3>canonicalizationAlgorithm</h3>
...
+        <h3>controller</h3>
         <p>
-An owner is an entity that claims control over a particular resource. Note
-that ownership is best validated as a two-way relationship where the owner
-claims ownership over a particular resource, and the resource clearly
-identifies its owner.
+A controller is an entity that claims control over a particular resource. Note
+that control is best validated as a two-way relationship where the controller
+claims control over a particular resource, and the resource clearly
+identifies its controller.
```
msporny, line 1712: We can't remove properties from this vocabulary because people may have deployed them. All we can do is deprecate properties. Please keep the section, `id`, and entry for for `owner` around... just note that it's deprecated and replaced by `controller`.

msporny (changes requested):
In general, +1 to the PR -- thank you @kkabrams for putting it together! The only change that we need to make is ensuring that we don't remove properties that might have deployment. `owner` has been around for close to a decade now, so we shouldn't remove it's entry, but rather, deprecate the entry. 

The whole vocabulary document could use fresh examples, see https://w3c.github.io/vc-data-model/#example-usage-of-the-id-property for an example (proof tab)... and DID Core for an example for how controller is used: https://www.w3.org/TR/did-core/#verification-methods ... but all that can be done in a separate PR.

kkabrams:
I looked through the other status values of properties and only saw "stable" and "unstable". The other deprecated properties were labeled as "unstable", but I wouldn't consider the owner property "unstable" since it's pretty old and will be stable in its "deprecated" state? I just have it marked as "deprecated" for now.

msporny (approved):

msporny:
> I just have it marked as "deprecated" for now.

Wonderful, and LGTM, thank you!

Issue #91, opened 2021-03-15
sec:authenticationMethod != sec:authentication


OR13:
https://github.com/w3c/did-spec-registries/issues/235

OR13:
@dlongley tell me what to do to fix this :)

iherman:
See https://github.com/w3c/did-spec-registries/issues/235#issuecomment-799511733

PR #128, opened 2021-12-18
Fix *Method definitions in namespace document

clehner:
Address #91.

- Change ids and IRIs for the following properties in the namespace document (web page) to correspond with the IRIs used in the Security Vocab and DID Core JSON-LD contexts:
  - authentication → authenticationMethod
  - capabilityDelegation → capabilityDelegationMethod
  - capabilityInvocation → capabilityInvocationMethod
  - keyAgreement → keyAgreementMethod
- Use both the short names (e.g. authentication) and the long names (e.g. authenticationMethod) in the headings. The short name is for consistency with use in JSON-LD documents with the existing context files, as an established convention; and the long name is for consistency with `assertionMethod` (which does not have this discrepancy) and with the IRI itself (which uses the long name as the suffix following the namespace part).
- ~~Add notes about how/why it is this way.~~

### DIFF
--- a/index.html
+++ b/index.html
@@ -1887,13 +1887,6 @@ to the identity <code>https://payswarm.example.com/i/bob</code>.
       typeof="rdf:Property">
         <h3>authentication (authenticationMethod)</h3>
 
-        <p class="note" title="Term differs from IRI suffix">
-For historical reasons, the term <code>authentication</code>
-is mapped to suffix <code>authenticationMethod</code> in the Security
-Vocabulary namespace, that is the following IRI:
-<code><a href="https://w3id.org/security#authenticationMethod">https://w3id.org/security#authenticationMethod</a></code>.
-        </p>
-
         <p>
           An authentication property is used to specify a URL that contains information about
           a verificationMethod used for authentication.
@@ -1939,13 +1932,6 @@ Vocabulary namespace, that is the following IRI:
       <section id="capabilityDelegationMethod" about="https://w3id.org/security#capabilityDelegationMethod"
       typeof="rdf:Property">
         <h3>capabilityDelegation (capabilityDelegationMethod)</h3>
-
-        <p class="note" title="Term differs from IRI suffix">
-For historical reasons, the term <code>capabilityDelegation</code>
-is mapped to suffix <code>capabilityDelegationMethod</code> in the Security
-Vocabulary namespace, that is the following IRI:
-<code><a href="https://w3id.org/security#capabilityDelegationMethod">https://w3id.org/security#capabilityDelegationMethod</a></code>.
-        </p>
 
         <p>
 A capabilityDelegation property is used to express that one or more
@@ -2010,13 +1996,6 @@ to the identity <code>did:example:123#WjKgJV7VRw3hmgU6--4v15c0Aewbcvat1BsRFTIqa5
       typeof="rdf:Property">
         <h3>capabilityInvocation (capabilityInvocationMethod)</h3>
 
-        <p class="note" title="Term differs from IRI suffix">
-For historical reasons, the term <code>capabilityInvocation</code>
-is mapped to suffix <code>capabilityInvocationMethod</code> in the Security
-Vocabulary namespace, that is the following IRI:
-<code><a href="https://w3id.org/security#capabilityInvocationMethod">https://w3id.org/security#capabilityInvocationMethod</a></code>.
-        </p>
-
         <p>
 A capabilityInvocation property is used to express that one or more
 verificationMethods are authorized to verify cryptographic proofs
@@ -2080,13 +2059,6 @@ to the identity <code>did:example:123#WjKgJV7VRw3hmgU6--4v15c0Aewbcvat1BsRFTIqa5
       <section id="keyAgreementMethod" about="https://w3id.org/security#keyAgreementMethod"
       typeof="rdf:Property">
         <h3>keyAgreement (keyAgreementMethod)</h3>
-
-        <p class="note" title="Term differs from IRI suffix">
-For historical reasons, the term <code>keyAgreement</code>
-is mapped to suffix <code>keyAgreementMethod</code> in the Security
-Vocabulary namespace, that is the following IRI:
-<code><a href="https://w3id.org/security#keyAgreementMethod">https://w3id.org/security#keyAgreementMethod</a></code>.
-        </p>
 
         <p>
           An keyAgreement property is used to specify a URL that contains information about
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1883,9 +1883,17 @@ <h3>assertionMethod</h3>
...
       typeof="rdf:Property">
-        <h3>authentication</h3>
+        <h3>authentication (authenticationMethod)</h3>
+
+        <p class="note" title="Term differs from IRI suffix">
+For historical reasons, the term <code>authentication</code>
+is mapped to suffix <code>authenticationMethod</code> in the Security
+Vocabulary namespace, that is the following IRI:
+<code><a href="https://w3id.org/security#authenticationMethod">https://w3id.org/security#authenticationMethod</a></code>.
+        </p>
```
msporny, line 1895: This presumes a tight coupling between DID Core and the Security Vocabulary. We shouldn't suggest that there is a tight coupling.

The Security Vocabulary doesn't need to make statements like this as how a JSON-LD context developer maps a term to a URL is entirely up to the context developer. That is, the Security Vocabulary could never know all the ways people map its URLs to terms. One developer may choose "authn" -> "https://w3id.org/security#authenticationMethod", while another chooses "验证" => "https://w3id.org/security#authenticationMethod" and another chooses "autenticación" => "https://w3id.org/security#authenticationMethod". All that to say, we shouldn't have this paragraph here. Sames goes for all the other entries.

If were were to add something to this effect somewhere, it should probably be in the DID Implementation Guide.

index.html
```diff
@@ -1928,9 +1936,17 @@ <h3>authentication</h3>
...
       typeof="rdf:Property">
-        <h3>capabilityDelegation</h3>
+        <h3>capabilityDelegation (capabilityDelegationMethod)</h3>
+
+        <p class="note" title="Term differs from IRI suffix">
+For historical reasons, the term <code>capabilityDelegation</code>
+is mapped to suffix <code>capabilityDelegationMethod</code> in the Security
+Vocabulary namespace, that is the following IRI:
+<code><a href="https://w3id.org/security#capabilityDelegationMethod">https://w3id.org/security#capabilityDelegationMethod</a></code>.
+        </p>
```
msporny, line 1948: Same as comment above, needs to be removed.

index.html
```diff
@@ -1990,9 +2006,17 @@ <h3>capabilityDelegation</h3>
...
       typeof="rdf:Property">
-        <h3>capabilityInvocation</h3>
+        <h3>capabilityInvocation (capabilityInvocationMethod)</h3>
+
+        <p class="note" title="Term differs from IRI suffix">
+For historical reasons, the term <code>capabilityInvocation</code>
+is mapped to suffix <code>capabilityInvocationMethod</code> in the Security
+Vocabulary namespace, that is the following IRI:
+<code><a href="https://w3id.org/security#capabilityInvocationMethod">https://w3id.org/security#capabilityInvocationMethod</a></code>.
+        </p>
```
msporny, line 2018: Same as above, needs to be removed.

index.html
```diff
@@ -2053,9 +2077,17 @@ <h3>capabilityInvocation</h3>
...
       typeof="rdf:Property">
-        <h3>keyAgreement</h3>
+        <h3>keyAgreement (keyAgreementMethod)</h3>
+
+        <p class="note" title="Term differs from IRI suffix">
+For historical reasons, the term <code>keyAgreement</code>
+is mapped to suffix <code>keyAgreementMethod</code> in the Security
+Vocabulary namespace, that is the following IRI:
+<code><a href="https://w3id.org/security#keyAgreementMethod">https://w3id.org/security#keyAgreementMethod</a></code>.
+        </p>
```
msporny, line 2089: Same as above, needs to be removed.

msporny (changes requested):
Minor change suggestion to remove explanation and place it elsewhere.

msporny:
@clehner You identified something that I'm sure is going to be confusing to most people. I'm going to document it here because I'm not sure where the explanation should go, since it has to do w/ a number of standards.

Waaay back in the day, probably close to 10 years ago at this point, the security vocab was created to implement what has become (Linked) Data Proofs/Integrity today. Before DID Documents existed, we stored this information on files on Web servers (the documents look close to what did:web looks like today). The concept of a "verification method" dates back that far, and a persons "DID Document" at the time would state how the person would digitally sign statements that originate from them. That's where "assertionMethod" comes from... "this is the verification method you should use when you are checking to see if I have asserted something about the world." That "assertion about something about the world" became what's known today as Verifiable Credentials... [blog post about that here](http://manu.sporny.org/2014/credential-based-login/).

Due to the [push back of one big company at W3C back in 2013](http://manu.sporny.org/2013/lds-vs-jose/), we had to delay the standardization of (Linked) Data Proofs/Integrity by roughly 7 years, which means that we ended up standardizing Verifiable Credentials before we standardized some of the proof mechanisms (which was backwards, but that's standards politics for you). In that 7 years, people wrote code, used the security vocabulary, deployed proof of concepts, refined their libraries and the term people used for the verification method that should be used to check statements that you said about the world ended up being "assertionMethod"... and that's what got locked in when the Verifiable Credentials data model was standardized:

https://github.com/w3c/vc-data-model/blob/v1.1/contexts/credentials/v1#L193-L194

```
            "assertionMethod": {"@id": "sec:assertionMethod", "@type": "@id", "@container": "@set"},
            "authentication": {"@id": "sec:authenticationMethod", "@type": "@id", "@container": "@set"}
```

Some think we really should've gone with "assertion", but it was a bit too late to change it at that point given all the software that had been written with the older term. We could change it in the future in Verifiable Credentials 2.0, but then everyone would have to update their software. The people using JSON-LD would have an easier time that those that don't... but then the problem becomes not just updating VC implementations, but all the signature suite libraries as well. At the time, it was deemed "not worth it", and so "assertionMethod" stuck around while the newer terms dropped the "Method" suffix in the JSON-LD Contexts.

It's a minor annoyance for those of us that like consistent terminology, and it is the unfortunate side-effect of standards politics delaying the (Linked) Data Proofs/Integrity work coupled with implementers implementing and deploying software in a way that makes it hard to change the values that they've shipped. Hope that helps explain why things are the way they are today.

OR13 (approved):
Looks like the feedback was implemented. @msporny would you consider adding a PR with the language examples you provided here?

msporny (approved):

msporny:
> Looks like the feedback was implemented. 

This is ready to merge. 

> @msporny would you consider adding a PR with the language examples you provided here?

I'm going to hold off on doing any of that since there is active debate on where such language should go:

https://github.com/w3c/did-spec-registries/pull/392#issuecomment-997432278

This PR should be merged immediately since it fixes an obvious bug in the document.

